PR #84, opened 2026-07-09
Sync from mono

onelxj:
Replaces the release-time snapshot with the current `public/` tree from the monorepo, as the source-of-truth seed for making `public/` a git submodule (vendored back into mono).

Merging this makes `reboot-dev/reboot@main` the content that mono will pin as the `public/` submodule.

The PR shows that all checks and the `public` related CI works well!


rjhuijsman (approved):
As discussed offline, this will of course need to be synced one more time to catch all of the most recent PRs - but excited to land this ASAP!

PR #92, opened 2026-07-15
Make agent not create `__init__.py` and add proper `.gitignore`

### LINKED ISSUE #91: Reboot skills improvements
**Describe the bug**
Skills create create workarounds / configurations that don't need to exist and don't `.gitignore` files I would expect to be ignored.

**To Reproduce**

These issues were discovered over a series of runs. 

Issus:
- [ ] Claude Code created unnecessary `__init__.py` files in /backend. We have a number of guards to prevent `__init__.py` files in /api but we have nothing to prohibit them in `/backend`. This doesn't cause an issue but they are unnecessary and the AI is likely to continue to create them because it is the standard way it expects packaging: a directory of modules should be an importable package.
- [ ] .rbt/ and generated file not added to `.gitignore`. We should add small .gitignore scaffolding that creates a `.gitignore` covering `.rbt/`, `backend/api`, `frontend/api`, .`venv`, `__pycache__`, `*.py[cod]`, `.mypy_cache/`, `node_modules`, `dist/` and `.venv`.

I will mention that I ran into a number of Envoy / Docker issues when doing this development. I traced this to problems with my environment 1. Claude Routines can only download packages from npm / pypi and envoy isn't availalbe there. 2. The other place I ran into issues was in GitHub Action Runners where envoy was not on my PATH by default and so it chose to use Docker instead, which greatly increased my test time. Not strictly an issue with skills though, juts an issue with the way I setup GitHub Action Runners.

I ran into the issue that commit db0d16de fixes: "Type error fix: drop Zod .default(undefined) on Optional fields" but that should be fixed upon next release.

**Impact**
How severely does this bug impact your use of Reboot?

- [ ] It's not having any negative impact
- [x] It's annoying
- [ ] It's a serious problem
- [ ] It makes it impossible for me to use Reboot

onelxj:
Fixes #91

### DIFF
--- a/reboot/plugin/skills/chat-app/references/project-shell.md
+++ b/reboot/plugin/skills/chat-app/references/project-shell.md
@@ -10,9 +10,11 @@ tags: project, shell, rbtrc, pyproject, python-version, main, application-entry,
 The `python` skill covers the canonical project shape
 (`lifecycle-project-setup.md`) and the `.rbtrc` format
 (`lifecycle-rbtrc.md`). That shape includes a project-root
-`.mypy.ini` (template in `lifecycle-project-setup.md`) — create it
-when scaffolding; it has no chat-app delta. What an MCP Chat App
-adds on top:
+`.mypy.ini` and a project-root `.gitignore` (templates in
+`lifecycle-project-setup.md`) — create both when scaffolding; they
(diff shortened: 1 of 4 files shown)


rileysdev (approved):
🙏

PR #99, opened 2026-07-20
OAuth: deliver identity provider claims to `User` state via `SetClaims`

rjhuijsman:
Previously, an auto-constructed `User` learned nothing about the person it represents: signing in through `oauth=` created the actor with a user ID and nothing else, and the verified identity information the identity provider holds (email, name, ... — the so-called "claims") were never available to the application. That's a problem for many applications that need things like a user's email or full name, for example also for Reboot Cloud.

This PR delivers verified identity provider claims to `User`. Application developers request the claims they want when constructing their identity provider:

```py
oauth=OAuthProviderByEnvironment(
    dev=Development(claims=["email"]),
    prod=Google(claims=["email"]),
),
```

Each provider declares which claims it can deliver and rejects any others at construction. When the developer requests a claim, the provider will add whatever OAuth scopes that claim needs automatically (e.g. `Google(claims=["email"])` adds the `email` scope). Claims can be remapped to a different name, for when your dev and prod providers disagree on what a claim is called (`claims={"email": "verified-email"}`). If no claims are requested, behavior is unchanged from what it was previously.

To use claims, application developers (if they want) override the implementation of a new `SetClaims`/`set_claims` transaction on their auto-constructed state type:

```py
# On `UserServicer`:
async def set_claims(
    self,
    context: TransactionContext,
    request: SetClaimsRequest,
) -> None:
    self.state.email = request.claims.get("email", "")
```

This approach mirrors the auto-constructor (`Create`/`create`) already present for `User`s, except that `create` is called exactly once (it's a constructor), whereas `set_claims` may be called many times (any time a claim may have changed).

Claims may have any type (string, bool, ...) so arrive as a plain `dict[str, Any]`, carried on the wire as a `map<string, google.protobuf.Value>`.

Calling `set_claims` is full-replace — "here is the complete, current set of verified claims" — and therefore idempotent, so if Reboot is unsure whether claims have changed it can simply rerun `set_claims`.

Calling `set_claims` happens on identity provider code exchange (every time a token is minted; the only moment fresh verified claims exist): first auto-construction's `create` happens (if needed), then `set_claims`. It's also possible (in future PRs) to hav webhooks call `set_claims`, so this design makes it possible for claims to update "reactively" too.

Please see individual commits for the steps needed to get there.

### DIFF
--- a/reboot/aio/auth/oauth_providers.py
+++ b/reboot/aio/auth/oauth_providers.py
@@ -19,7 +19,7 @@ from reboot.crypto import root_keys
 from reboot.run_environments import running_rbt_dev
 from starlette.requests import Request
 from starlette.responses import HTMLResponse
-from typing import Any, NewType, Optional
+from typing import Any, Mapping, NewType, Optional, Sequence
 from ulid import ULID
 from urllib.parse import urlencode, urlparse, urlunparse
 
@@ -41,6 +41,18 @@ class ExchangeResult:
     # constructed with `store_tokens=True`; `None` otherwise, so apps
     # that don't opt in never carry these secrets around.
     tokens: Optional[OAuthTokens]
+    # The user's verified identity claims — the claims the provider
+    # was constructed to deliver (`claims=`), keyed by their presented
+    # names, as JSON-serializable values. Only claims from a verified
+    # source belong here: an ID token received directly from the
+    # provider's token endpoint over TLS, or the provider's userinfo
+    # endpoint over TLS. A mapping — even an empty one — is the
+    # complete, current verified claim set and is delivered as a full
+    # replacement, so an empty mapping clears previously delivered
+    # claims. `None` means claim delivery is off — the provider was
+    # constructed without `claims=` or has no verified claims source —
+    # and delivers nothing.
+    claims: Optional[Mapping[str, Any]] = None
 
 
 def _expires_at_from_expires_in(expires_in: Any) -> Optional[int]:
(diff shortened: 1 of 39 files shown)


reboot/aio/auth/oauth_providers.py
```diff
@@ -87,14 +99,130 @@ class OAuthProvider(ABC):
     Base class for identity providers.
     """
 
+    # The identity claims this provider can deliver: each available
+    # claim name, mapped to the OAuth scope that makes the identity
+    # provider include it (`None` when no scope is needed). Subclasses
+    # with a verified claims source override this.
+    _AVAILABLE_CLAIMS: Mapping[str, Optional[str]] = {}
```
onelxj, line 106: Shall we make it configurable at construction time instead of overrides?
We expose `Google`, `GitHub` and `Auth0` to be configurable so if somebody wants more items in the `claims` they will need to override our OAuth provider class now.

reboot/pydantic_schema_to_proto.py
```diff
@@ -486,6 +502,11 @@ async def generate(
                 type_name = "double"
             elif value_type == bool:
                 type_name = "bool"
+            elif value_type is Any:
```
onelxj, line 518: I'm not entirely sure that if somebody will try to annotate the field as `list[Any]` or just a simple `Any` this all will work. Let's add some failure-tests or make sure we support that.

reboot/templates/reboot.py.j2
```diff
@@ -4069,12 +4115,17 @@ class {{ state.proto.name }}Servicer({{ state.proto.name }}BaseServicer):
...
-    {% if method.proto.name == AUTO_CONSTRUCT_PROTO_METHOD and state.proto.auto_construct != AUTO_CONSTRUCT_UNSPECIFIED %}
+    {% if state.proto.auto_construct != AUTO_CONSTRUCT_UNSPECIFIED and method.proto.name in [AUTO_CONSTRUCT_PROTO_METHOD, SET_CLAIMS_PROTO_METHOD] %}
+    {% if method.proto.name == AUTO_CONSTRUCT_PROTO_METHOD %}
     # Concrete no-op default for `{{ AUTO_CONSTRUCT_METHOD }}`. The
     # auto-constructed `{{ AUTO_CONSTRUCT_METHOD }}` is always a
     # `Transaction` (see `api.py`), so it takes a `TransactionContext`.
     # Override in your Servicer to run custom initialization
     # on new {{ state.proto.name }} instances.
+    {% else %}
+    # Concrete no-op default for `{{ SET_CLAIMS_METHOD }}`. You may
```
onelxj: Do we really want a no-op default if somebody explicitly specified `claims` instead of an error?

onelxj (approved):

reboot/aio/auth/oauth_providers.py
rjhuijsman: I think we already have what you're asking for; but let me make sure:
* the `_AVAILABLE_CLAIMS` is a list of claims the identity provider _can_ provide. That list is essentially just documentation (it's out of our control what claims e.g. Google will give you); it's here to catch typos early.
* The `claims` parameter (e.g. `Google([...], claims=["claims", "that", "I", "want"])`) lets the developer choose the claims they'd like to actually get (choosing from the menu of `_AVAILABLE_CLAIMS`).
* Therefore they'll never override an existing auth provider: they can't reasonably expand the menu (unless they're themselves Google), and they can freely change their choices from the menu by passing a parameter.

Does that make sense?

reboot/templates/reboot.py.j2
rjhuijsman: Good question... We could indeed fail `SetClaims` and thereby the login if `set_claims` wasn't overridden... If we provided good enough error messages (in the server logs and perhaps e.g. the frontend's `console.log`) then that might be helpful. Let me try it out!

reboot/aio/auth/oauth_providers.py
onelxj: For `Google` the menu right now is

```
 _AVAILABLE_CLAIMS = {
        "email": "email",
        "email_verified": "email",
        "name": "profile",
    }
```

my robot says that the list might look like that:

```
_AVAILABLE_CLAIMS = {
    "email": "email",
    "email_verified": "email",
    "name": "profile",
    "given_name": "profile",
    "family_name": "profile",
    "picture": "profile",
    "profile": "profile",   # URL of the user's profile page (rare, mostly legacy G+)
    "locale": "profile",    # deprecated, often absent now
}
```

So we are missing a couple of them for now.

reboot/pydantic_schema_to_proto.py
onelxj: There is a specific `Json` Pydantic type:
https://pydantic.dev/docs/validation/2.0/usage/types/json/

Shall we use it instead? Maybe as `Json[Any]`?

reboot/templates/reboot.py.j2
rjhuijsman: Done.

reboot/pydantic_schema_to_proto.py
rjhuijsman: Indeed `list[Any]` and plain `Any` weren't supported - but that was a trivial fix, so done!

I think I prefer `dict[str, Any]` here over `Json[Any]`, for a few reasons:
1. `Json` implies that your code will get a `dict[str, Any]`, if you literally write `dict[str, Any]` that's spelled out more clearly.
2. If we use `Json` it means we're shipping a JSON string on the wire, which not only is less efficient and correct than proto's `map<string, Value>`, but it also would need considerable TypeScript generated code surgery to make it work for TS (backend) clients. I don't think that work is worth it right now, because...
3. I don't see any advantages for `Json` in this situation.

reboot/aio/auth/oauth_providers.py
rjhuijsman: Good catch; audited all three providers, gave the complete menus for both, and added links to the relevant docs.

PR #104, opened 2026-07-21
OAuth: add Ory support, including "reactive" claims

rjhuijsman:
This PR adds an `Ory` OAuth provider (what Reboot Cloud will use).

This PR also adds the mechanics needed to support "reactive" claims: Ory can send a webhook when claims change (e.g. a user changes their email address), and we now support receiving that webhook and immediately updating claims.

Finally, because Reboot Cloud is migrating from a hand-rolled `User`-construction story to "proper" `User`-auto-construction, we must support cases where the `User` being auto-constructed was actually already constructed by a different mechanism earlier. This PR adds a commit for that edge case.

Please see individual commits!

### DIFF
--- a/reboot/aio/auth/oauth_providers.py
+++ b/reboot/aio/auth/oauth_providers.py
@@ -76,6 +76,39 @@ def _scopes_from_response(granted: Any, requested: list[str]) -> list[str]:
     return requested
 
 
+def _normalize_domain(domain: Optional[str]) -> str:
+    """Strip a leading scheme and any trailing slash from a
+    provider's `domain` so both `tenant.example.com` and the full
+    `https://tenant.example.com/` form work; returns `""` for a
+    missing domain (caught later by `validate()`).
+    """
+    if not domain:
+        return ""
+    parsed = urlparse(domain)
+    # `urlparse("tenant.example.com")` puts everything in `path` (no
+    # scheme), whereas `urlparse("https://tenant.example.com")` puts
+    # it in `netloc`; take whichever is populated.
+    host = parsed.netloc or parsed.path
+    return host.strip("/")
+
+
+def _decoded_id_token(
+    id_token: Optional[str],
+) -> Optional[dict[str, Any]]:
+    """Decode an OIDC ID token's claims, or `None` if there's no
+    token.
+
+    Decodes without verifying the signature: the token was received
+    directly from the provider's token endpoint over TLS, so the
+    transport guarantees authenticity (the same reasoning `Google`
+    applies). We don't restrict `algorithms` because the signature
+    isn't checked and providers may sign with RS256 or HS256.
+    """
+    if not id_token:
+        return None
+    return jwt.decode(id_token, options={"verify_signature": False})
+
+
 def origin_from_request(request: Request) -> str:
     """
     Derive the server's origin (`scheme://host`) from request headers.
@@ -789,40 +822,7 @@ class Auth0(RegisteredOAuthProvider):
             store_tokens=store_tokens,
             claims=claims,
         )
-        self._domain = self._normalize_domain(domain)
-
-    @staticmethod
-    def _normalize_domain(domain: Optional[str]) -> str:
-        """Strip a leading scheme and any trailing slash from an Auth0
-        `domain` so both `your-tenant.auth0.com` and the full
-        `https://your-tenant.auth0.com/` form work; returns `""` for a
-        missing domain (caught later by `validate()`).
-        """
-        if not domain:
-            return ""
-        parsed = urlparse(domain)
-        # `urlparse("tenant.auth0.com")` puts everything in `path` (no
-        # scheme), whereas `urlparse("https://tenant.auth0.com")` puts it
-        # in `netloc`; take whichever is populated.
-        host = parsed.netloc or parsed.path
-        return host.strip("/")
-
-    @staticmethod
-    def _decoded_id_token(
-        id_token: Optional[str],
-    ) -> Optional[dict[str, Any]]:
-        """Decode an OIDC ID token's claims, or `None` if there's no
-        token.
-
-        Decodes without verifying the signature: we received this token
-        directly from the provider's token endpoint over TLS, so the
-        transport guarantees authenticity (the same reasoning `Google`
-        applies). We don't restrict `algorithms` because the signature
-        isn't checked and Auth0 tenants may sign with RS256 or HS256.
-        """
-        if not id_token:
-            return None
-        return jwt.decode(id_token, options={"verify_signature": False})
+        self._domain = _normalize_domain(domain)
 
     @property
     def _authorization_endpoint(self) -> str:
@@ -897,7 +897,7 @@ class Auth0(RegisteredOAuthProvider):
                 response.raise_for_status()
                 token_response = await response.json()
 
-            decoded = self._decoded_id_token(token_response.get("id_token"))
+            decoded = _decoded_id_token(token_response.get("id_token"))
             user_id = decoded.get("sub") if decoded is not None else None
             # The ID token carries the requested identity claims (we
             # requested the scopes they need); surface whichever are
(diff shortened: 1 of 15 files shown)


reboot/aio/auth/oauth_providers.py
```diff
@@ -964,6 +964,206 @@ def _tokens_from_response(
...
+    `your-slug.projects.oryapis.com`), so like `Auth0` this provider
+    takes a `domain`. Create an OAuth2 client in the Ory project with
+    the authorization-code grant, add Reboot's callback
+    (`<base>/__/oauth/callback`) to its redirect URIs, and pass the
+    client's ID/secret here.
+
+    Obtains the user ID from the OIDC ID token's `sub` claim — for
+    Ory that is the Kratos identity ID (assuming the default,
+    non-pairwise subject configuration) — falling back to the
+    `/userinfo` endpoint if no ID token is returned. Pass `claims=`
```
rjhuijsman: Under what situation does Ory not deliver an ID token? Do we really need the fallback logic?

reboot/aio/auth/oauth_providers.py
```diff
@@ -964,6 +964,206 @@ def _tokens_from_response(
...
+    # `sub` claim (the Kratos identity ID).
+    _REQUIRED_SCOPE = "openid"
+
+    # The standard OIDC identity claims Ory maps the identity's
+    # traits into, with the scope that makes each appear.
+    _AVAILABLE_CLAIMS = {
+        "email": "email",
+        "email_verified": "email",
+        "name": "profile",
+    }
```
rjhuijsman: Are these all of the scopes Ory offers? Please link to the documentation of the available OpenID scopes like we do for the other providers.

reboot/aio/auth/oauth_providers.py
```diff
@@ -964,6 +964,206 @@ def _tokens_from_response(
...
+    def token_service_id(self) -> str:
+        # Ory serves one project per domain, so the project domain
+        # (e.g. "your-slug.projects.oryapis.com") names the service.
+        return self._domain
+
+    def validate(self) -> None:
+        # Validate the credentials the base class knows about, then
+        # the `domain` that's unique to Ory — same fail-fast contract,
+        # so a missing domain is caught at selection time, not
+        # mid-sign-in.
```
rjhuijsman: This comment seems a bit useless, remove it.

reboot/aio/auth/oauth_providers.py
```diff
@@ -964,6 +964,206 @@ def _tokens_from_response(
...
+                    self._userinfo_endpoint,
+                    headers={"Authorization": f"Bearer {access_token}"},
+                ) as response:
+                    response.raise_for_status()
+                    user_info = await response.json()
+                user_id = user_info.get("sub")
+                # `/userinfo` responses use the same standard claim
+                # names as the ID token, and the TLS round-trip to
+                # Ory makes them a verified source too.
+                claims = self._presented_claims(user_info)
```
rjhuijsman: This is a lot of complication for something that shouldn't happen. If we don't have a strong reason for it, I'd prefer to remove it.

reboot/aio/auth/oauth_server.py
```diff
@@ -322,6 +323,7 @@ def __init__(
         authenticated: Optional[
             Callable[[ExternalContext, str, Optional[Mapping[str, Any]]],
                      Awaitable[None]]] = None,
+        set_claims_if_exists: Optional[SetClaimsIfExists] = None,
```
rjhuijsman: This is a sibling of `authenticated`, which is named after _the moment the hook should be called_, not after _what the hook does_. Let's follow that pattern, e.g.: `claims_changed`.

reboot/aio/auth/oauth_providers.py
```diff
@@ -1015,6 +1098,10 @@ def __init__(
         scopes: Optional[list[str]] = None,
         store_tokens: bool = False,
         claims: Optional[Sequence[str] | Mapping[str, str]] = None,
+        # Shared secret authenticating calls to the settings-flow
+        # webhook (see the class docstring); the webhook route is
+        # only mounted when one is set.
+        webhook_secret: Optional[str] = None,
```
rjhuijsman, line 1120: Let's instead compute a webhook secret the same way that we compute the token signing secret?

reboot/templates/reboot.py.j2
```diff
@@ -3138,9 +3138,22 @@ class {{ state.proto.name }}BaseServicer(IMPORT_reboot_aio_servicers.Servicer):
...
         if claims is not None:
+{% if options.proto.pydantic %}
             await {{ state.proto.name }}.ref(state_id).always().{{ SET_CLAIMS_METHOD }}(
                 context, claims=claims
             )
+{% else %}
+            # This state's `{{ SET_CLAIMS_METHOD }}` request carries
+            # claims as a `map<string, google.protobuf.Value>`, but the
+            # framework delivers them as native Python values; round
+            # them through a `Struct`, which wraps each native value in
```
rjhuijsman: "round them"..? Pick a clearer way of expressing this.

reboot/aio/auth/oauth_providers.py
rjhuijsman: 👍

reboot/aio/auth/oauth_providers.py
rjhuijsman: 👍

reboot/aio/auth/oauth_providers.py
rjhuijsman: 👍

reboot/aio/auth/oauth_providers.py
rjhuijsman: 👍

reboot/aio/auth/oauth_server.py
rjhuijsman: 👍

reboot/aio/auth/oauth_providers.py
rjhuijsman: 👍

reboot/templates/reboot.py.j2
rjhuijsman: 👍

reboot/aio/auth/oauth_providers.py
```diff
@@ -984,6 +1032,41 @@ class Ory(RegisteredOAuthProvider):
...
+    otherwise only propagate that change to the application at their
+    next sign-in. Pass a `webhook_secret` (with `claims=`, since
+    delivering claims is all the webhook does) and this provider
+    exposes `POST /__/oauth/ory/webhook`; configure an Ory Action to
+    call it after the settings flow and identity changes propagate
+    immediately — but only for users who have already signed into
+    this application (the webhook never materializes a `User` for an
+    identity that has not). In the Ory project config, under
+    `selfservice.flows.settings.after`:
+
```
rjhuijsman: If there's documentation you can link to for how to do this in the Ory dashboard, please link it.

rjhuijsman: 👍

reboot/aio/auth/oauth_providers.py
```diff
@@ -964,6 +964,180 @@ def _tokens_from_response(
...
+    # `sub` claim (the Kratos identity ID).
+    _REQUIRED_SCOPE = "openid"
+
+    # The identity claims Ory maps the identity's traits into, each
+    # keyed to the standard OIDC scope that makes it appear: `email` /
+    # `email_verified` need the `email` scope, `name` the `profile`
+    # scope. Ory (Hydra) exposes the standard OIDC scopes and their
+    # claims — nothing Ory-specific — so this is the menu; absent
+    # claims are simply omitted (see `_presented_claims`). See
+    # https://openid.net/specs/openid-connect-core-1_0.html#ScopeClaims.
```
onelxj: There are much more claims under that link, why didn't you specify all of them there?

onelxj (approved):

reboot/aio/auth/oauth_providers.py
```diff
@@ -964,6 +1002,364 @@ def _tokens_from_response(
...
+                reason=(
+                    "`Ory` got a `webhook_secret=` but no `claims=`: "
+                    "the settings-flow webhook exists to deliver "
+                    "identity claims, so pass `claims=` (e.g. "
+                    "`claims=['email']`) or remove the "
+                    "`webhook_secret=`."
+                ),
+            )
+        self._domain = _normalize_domain(domain)
+        self._webhook_secret = webhook_secret
```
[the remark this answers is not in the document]
rjhuijsman: 👍 Fixed. `Ory.validate()` now refuses an empty `webhook_secret` — an empty secret authenticates nobody, so it's a misconfiguration rather than an off switch (`webhook_secret=None`, the default, is what leaves the webhook unexposed). `validate()` runs at provider selection, i.e. before `mount_routes`, so this fails fast instead of quietly leaving the route unmounted. Covered by `OryWebhookSecretRequiresClaimsTest.test_empty_webhook_secret_rejected`.

reboot/aio/auth/oauth_providers.py
```diff
@@ -964,6 +1002,364 @@ def _tokens_from_response(
...
+            )
+
+        body = await request.body()
+        try:
+            payload = json.loads(body)
+        except json.JSONDecodeError:
+            return JSONResponse(
+                {"error": "invalid_json"},
+                status_code=400,
+            )
```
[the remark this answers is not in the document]
rjhuijsman: 👍 Fixed. The `except` now covers `UnicodeDecodeError` alongside `json.JSONDecodeError`, so a non-UTF-8 body is a 400 rather than a 500. `OryWebhookTest.test_malformed_body_is_a_bad_request` posts both unparsable JSON and a body that isn't valid UTF-8 and asserts the 400.

reboot/aio/auth/oauth_providers.py
```diff
@@ -964,6 +1002,364 @@ def _tokens_from_response(
...
+        # exists: an identity that never signed into this application
+        # is skipped (Ory webhooks fire for every identity in the
+        # project), and it converges on the delivered claims for one
+        # that has. Retries and replays are safe — `set_claims` is a
+        # full replace, idempotent by contract.
+        await self._claims_changed(
+            external_context(request),
+            UserId(identity_id),
+            claims,
+        )
```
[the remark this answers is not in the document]
rjhuijsman: 👍 for the read, but leaving this as is. `UserId = NewType("UserId", str)`, so a `UserId` *is* a `str` to type checkers — this passes the mypy aspect as written. It's also the established pattern in this file: `OAuthServer._mint_tokens_for_user` takes a `UserId` and hands it to the `authenticated` callback, which is likewise typed `str`. Narrowing only `ClaimsChanged` would make it diverge from its sibling callback for no type-safety gain.

reboot/aio/auth/oauth_providers.py
rjhuijsman: After really fighting Claude about it, we now have the _actual_ full list of Ory claims.

reboot/templates/reboot.py.j2
```diff
@@ -3125,9 +3138,70 @@ class {{ state.proto.name }}BaseServicer(IMPORT_reboot_aio_servicers.Servicer):
...
+            # A `{{ SET_CLAIMS_METHOD }}` on a state that was never
+            # constructed aborts with `StateNotConstructed` (and logs
+            # one `WARNING`); that is exactly the "identity never
+            # signed in here" case, so swallow it. Any other abort is
+            # a real error and propagates.
+            if isinstance(
+                aborted.error,
+                IMPORT_rbt_v1alpha1.errors_pb2.StateNotConstructed,
+            ):
+                return
```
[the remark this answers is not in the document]
rjhuijsman: Such a big change to swallow an occasional warning isn't worth the work right now.

PR #105, opened 2026-07-21
Migrate constructor/factory methods to `MyType.factory.my_method`

### LINKED ISSUE #106: Constructor/factory methods collide with type-class-level methods
We currently have two overlapping problems:
1. If I add a constructor/factory method named `ref`, it'll silently collide with the `MyStateType.ref()` method that's already present on every state type class. We currently have nothing that prevents that, and it almost certainly surfaces as a nasty type error or runtime crash.
2. Even if we detected current naming collisions and prevented them, the fact that constructors/factories are on `MyStateType` prevents us from _ever safely adding any new methods to that state type class_.
  * For example, if I wanted to add `MyStateType.query(...)` to the framework, I couldn't do that safely: a customer may already have declared a `query` constructor/factory.
  * If I added `query` anyway, any customer with a `query` constructor/factory would now be permanently stuck, unable to upgrade: they can't rename/delete their factory method (Reboot doesn't support that, yet).

The conclusion from my brainstorm with @benh, even though neither of us _love_ it: let's namespace all factory methods under `MyStateType.factory`, so e.g. `await MyStateType.create(context)` becomes `await MyStateType.factory.create(context)`. That frees up the top-level namespace for any future additions we might want to make. We're OK with the slighly-less-nice-to-write code because it's something agents will write, and it still reads pretty clearly to a human.

We should make this change, and do a thorough pass through the skills in `public/reboot/plugin` and the examples in `public/reboot/examples` to match this new pattern.

rjhuijsman:
Fixes https://github.com/reboot-dev/reboot/issues/106

Before this change, constructors lived directly on state classes (e.g. `Account.open()`). A user-defined constructor named `ref` would have silently collided with the built-in `MyType.ref()`, and the framework could never add a new factory method without risking a clash with an existing user constructor — which a customer also can't rename or delete, leaving them permanently unable to upgrade.

This PR namespaces every state type's constructor (a.k.a. factory) methods behind a `factory()` function, so `await MyType.create(context)` becomes `await MyType.factory().create(context)` — in Python, TypeScript and the web client alike. That frees the top level of the generated state class for framework methods we may want to add later (e.g. a future `MyType.query(...)`), and prevents naming collisions with built-in methods. `factory()` is a function rather than an attribute so that we can give it more to do later (say, return an instance that carries per-call state) without another migration of every caller; for now it simply returns the `_Factory` class. How a constructor is _declared_ (`factory=True`, or the `constructor` proto option) is unchanged — only the call surface moved.

**Reviewer hint:** the GitHub diff of the `*.j2` template files looks big, but very little actually changed — it's ~all mechanical. In Python the constructor and idempotency-scope methods are simply indented into a new nested `class _Factory` (literally +4 spaces on everything). In TypeScript the same methods had to be moved to keep them contiguous. No control flow or logic changed in either language. The only "functional" changes are a new class to wrap these methods, a `factory()` accessor on the state class, and a `.factory()` in places they're called. The `*.golden.*` files are regenerated codegen output.

The last two commits respond to the review: one makes `factory` a function (it started out as an attribute), the other audits everything that landed on `main` since this PR was cut for dependencies on the old shape — the dashboard's code watcher, the BDD fixtures, and the new callers — and moves them over.

I suggest you review the commits individually.


rjhuijsman:
@benh poke!

rjhuijsman:
@benh poke again! Let's get this landed!

reboot/templates/reboot.py.j2
```diff
@@ -7121,268 +7121,271 @@ class {{ client.proto.state_name }}:
...
+            return cls.idempotently(
+                how=IMPORT_reboot_aio_idempotency.ALWAYS,
+            )
 
-    @classmethod
-    def always(cls):
-        return cls.idempotently(
-            how=IMPORT_reboot_aio_idempotency.ALWAYS,
-        )
+    factory = _Factory
```
benh, line 7388: My only request is to make this a function so that if we ever find ourselves wanting to provide more functionality we can do so. I _think_ that means we can keep returning `_Factory` from that function and that if in the future we instead end up creating instances of `_Factory` then the only code that would have to change are folks that stored `<StateType>._Factory` but I don't think that should be anyone and that is an easy thing for coding agents to do as part of a release migration.

benh (approved):

benh (changes requested):
Shoot I lied, one more thing that needs to get done is to ask your coding agent to update any of the static analysis that we've since introduced as part of the dashboard for tracking factory calls. Better yet, just ask it to review all commits since this PR was created to confirm that there aren't any other change that depend on the previous semantics of factories/constructors. Thanks!

reboot/templates/reboot.py.j2
rjhuijsman: Done. 👍🏻

PR #109, opened 2026-07-23
Blob storage: `Blob` state machine with a pluggable gRPC data plane

rjhuijsman:
Before this change, Reboot had no first-class way to store large binary objects: state machines hold protobuf state, which is unsuited to multi-megabyte payloads, so applications had nowhere to put user uploads like images or videos.

This PR introduces the `Blob` state type as the **control plane** for one immutable-once-committed binary object: its state holds only metadata (content type, size, upload progress, lifecycle, authorization) while the bytes travel directly between the client and a **data plane**. Data is uploaded to the data plane in parts, each via `PUT` to a URL minted by the control plane - compatible with S3 as a data plane, if we ever want that. Uploads are resumable, sizes are enforced against the real bytes at commit time, and never-committed blobs expire automatically.

The data plane is a **plain gRPC service, `BlobDataPlane`** (`data_plane.proto` — no Reboot options, implementable by anything), discovered via `REBOOT_BLOB_DATA_PLANE_URL`. Locally (`rbt dev run` / `rbt serve run` / unit tests) a filesystem-based data plane server is run out-of-the-box.

Authorization: `Blob` has two mechanisms that combine:
1. `Blob` creation is always application-mediated; it can't be done directly from a frontend, creation must go through a backend call. The backend can then use whatever existing auth mechanism they'd like. Assuming the `Blob` is given a random ID, knowing that state ID acts as a capability: to upload or download you must first know the ID.
2. `Blob`s can be created with an `uploader_id` and `downloader_ids`; if given these will limit uploads and downloads to only users whose IDs are in those lists.

Reviewer hint: review commits in-order.

TESTED: with new unit tests, and by manually running `rbt dev run`.


rbt/std/blobs/v1/blobs.proto
```diff
@@ -0,0 +1,414 @@
...
+message NotCommitted {}
+
+// Raised when the uploaded parts violate the blob's declared `size`
+// or `max_size`.
+message SizeMismatch {
+  // The total size of all parts reported uploaded so far.
+  uint64 bytes_uploaded = 1;
+}
+
+// Raised when `Commit` is called with no parts uploaded, or with a
```
rjhuijsman: "no parts", or "too few parts"?

rbt/std/blob/v1/blob.proto
```diff
@@ -0,0 +1,414 @@
...
+    UPLOADING = 0;
+
+    // `Commit` was called; the `CompleteUpload` workflow is
+    // finalizing the object in the storage backend.
+    COMMITTING = 1;
+
+    // The blob's bytes are immutable and downloadable.
+    COMMITTED = 2;
+
+    // `Remove` was called (or the upload expired); the
```
rjhuijsman, line 79: The name of the method (`Remove`) doesn't match the name of the status (`DELETING`), which is confusing. Let's make the two match; I have a slight preference for renaming the method to `Delete` if possible.

rbt/std/blobs/v1/blobs.proto
```diff
@@ -0,0 +1,414 @@
...
+  option (rbt.v1alpha1.state) = {
+    trusted_effects: true,
+  };
+
+  enum Status {
+    // Parts may be uploaded. The initial status.
+    UPLOADING = 0;
+
+    // `Commit` was called; the `CompleteUpload` workflow is
+    // finalizing the object in the storage backend.
```
rjhuijsman: Here we say "storage backend", elsewhere we say "data plane". Let's standardize on "data plane".

rbt/std/blob/v1/blob.proto
```diff
@@ -0,0 +1,414 @@
...
+  // MIME type served on download, e.g. `image/png`.
+  string content_type = 2;
+
+  // Expected total size in bytes, when known at `Create` time;
+  // enables definite progress tracking and is enforced at `Commit`.
+  optional uint64 size = 3;
+
+  // Upper bound on the total size in bytes, enforced as parts are
+  // reported and at `Commit`.
+  optional uint64 max_size = 4;
```
rjhuijsman: Mutually exclusive with `size`? Therefore possibly a `oneof`?

rbt/std/blobs/v1/blobs.proto
```diff
@@ -0,0 +1,414 @@
...
+
+  // Expected total size in bytes, when known at `Create` time;
+  // enables definite progress tracking and is enforced at `Commit`.
+  optional uint64 size = 3;
+
+  // Upper bound on the total size in bytes, enforced as parts are
+  // reported and at `Commit`.
+  optional uint64 max_size = 4;
+
+  // The user that may upload into this blob. Empty means anyone who
```
rjhuijsman: The _ID of the_ user.

rbt/std/blobs/v1/blobs.proto
```diff
@@ -0,0 +1,414 @@
...
+  // knows this blob's id may upload; see the authorizer note in
```
rjhuijsman: id -> ID; here and everywhere in comments.

rbt/std/blobs/v1/blobs.proto
```diff
@@ -0,0 +1,414 @@
...
+  // `reboot.std.blobs.v1.blobs`.
+  string uploader_id = 5;
+
+  // Who may download this blob. When unset (omitted at `Create`),
```
rjhuijsman: "IDs of users who may [...]"

rbt/std/blobs/v1/blobs.proto
```diff
@@ -0,0 +1,414 @@
...
+  // later with `SetDownloaders`.
+  optional Downloaders downloader_ids = 5;
+}
+
+message CreateResponse {}
+
+////////////////////////////////////////////////////////////////////////
+
+message SetDownloadersRequest {
+  // The new download allow-list. Omit the field entirely to remove
```
rjhuijsman: "new"? If this is referring to this list being new, consult your rules against diff-style comments. If this is referring to "new downloads", please clarify to avoid this confusion.

rbt/std/blob/v1/blob.proto
rjhuijsman, line 1: If we compare to other API directories in `rbt/v1alpha1` or `rbt/std`, none of them use plural names. Let's match that: `rbt/std/blob/v1/blob.proto`.

rbt/std/blobs/v1/data_plane.proto
```diff
@@ -0,0 +1,186 @@
...
+// objects, and the bytes themselves travel directly between the client
+// and wherever the data plane keeps them (a presigned-URL store like
+// S3, or — for a data plane whose `Configuration` asks for forwarded
+// paths — the application's own HTTP routes proxying to it).
+//
+// This is a plain gRPC service, deliberately free of any Reboot
+// framework options, so that it can be implemented by anything and
+// addressed by a bare URL (`REBOOT_BLOB_DATA_PLANE_URL`). The
+// open-source filesystem server implements it; Reboot Cloud implements
+// it privately (S3, and eventually a CDN) in each app's facilitator.
```
rjhuijsman: Drop this comment; if we add more/different implementations it's stale, and it references things that aren't even in this repo (so totally can't be kept in sync, nor are any of its concern).

rbt/std/blobs/v1/data_plane.proto
```diff
@@ -0,0 +1,186 @@
...
+
+service BlobDataPlane {
+  // Reports what the control plane needs to know to talk to this data
+  // plane: the client part size, and which paths (if any) the
+  // application must forward to it. Read once at application startup.
+  rpc Configuration(ConfigurationRequest) returns (ConfigurationResponse);
+
+  // Provisions an upload session for a blob and returns its data-plane
+  // upload id (e.g. an S3 multipart upload id). Should reuse an
+  // existing uncommitted session for the same blob; a fresh session is
```
rjhuijsman: "for the same blob where possible; a [...]"

rbt/std/blobs/v1/data_plane.proto
```diff
@@ -0,0 +1,186 @@
...
+  // workflow retries. Idempotent: completing an already-completed blob
+  // returns its ETag.
+  rpc CompleteUpload(DataPlaneCompleteUploadRequest)
+      returns (DataPlaneCompleteUploadResponse);
+
+  // Mints a URL from which the committed object's bytes can be
+  // downloaded: absolute (e.g. a presigned S3/CloudFront URL) or
+  // application-relative under a forwarded path, like
+  // `UploadInstructions` URLs. Minted afresh per request so that
+  // swapping the data-plane implementation takes effect immediately.
```
rjhuijsman: "Minted afresh" is an implementation detail that depends on both the control plane and the data plane; it probably shouldn't be part of the API specification.

rbt/std/blob/v1/blob.proto
rjhuijsman: 👍🏻

rbt/std/blob/v1/blob.proto
rjhuijsman: If we know `size`, then when is it useful to have a `max_size` that's not exactly `size`?
* If `max_size` is smaller: bug, upload can't complete.
* If `max_size` is bigger: allows uploads to grow bigger than what will be accepted at commit.

So if `size` is set, it seems like we should also use that value as our `max_size`, and we shouldn't permit `max_size` to also be set.

reboot/std/blob/v1/_data_plane.py
```diff
@@ -0,0 +1,93 @@
+"""Client-side glue for talking to a `BlobDataPlane` gRPC service.
+
+The `Blob` control-plane servicer holds no bytes; it calls a data-plane
+service (discovered via `REBOOT_BLOB_DATA_PLANE_URL`) to provision
+uploads, finalize them, mint URLs, and delete objects. Every local run
+mode provides a filesystem data plane automatically: `rbt dev run` and
+`rbt serve run` spawn it as a subprocess, and the
+`reboot.aio.tests.Reboot` harness runs it in-process. Deployments set
+the URL themselves (Reboot Cloud points it at the app's facilitator).
```
rjhuijsman: All of this comment is about code that's not in this file. Remove it.

reboot/std/blob/v1/blob.py
```diff
@@ -0,0 +1,625 @@
...
+    exact `size` is its own ceiling: bytes beyond it could never be
+    committed, so there is no reason to accept them."""
+    if state.HasField("size"):
+        return state.size
+    if state.HasField("max_size"):
+        return state.max_size
+    return None
+
+
+def _uploader_or_open(*, context, state=None, request=None, **kwargs):
```
rjhuijsman: Please type the return type

reboot/std/blob/v1/blob.py
```diff
@@ -0,0 +1,625 @@
...
+    if state.uploader_id == "":
+        return rbt.v1alpha1.errors_pb2.Ok()
+    if context.auth is None or context.auth.user_id is None:
+        return rbt.v1alpha1.errors_pb2.Unauthenticated()
+    if context.auth.user_id == state.uploader_id:
+        return rbt.v1alpha1.errors_pb2.Ok()
+    return rbt.v1alpha1.errors_pb2.PermissionDenied()
+
+
+def _downloader_or_open(*, context, state=None, request=None, **kwargs):
```
rjhuijsman: Please type the return type

reboot/std/blob/v1/blob.py
```diff
@@ -0,0 +1,625 @@
...
+
+
+class BlobServicer(Blob.Servicer):
+
+    # The data-plane gRPC stub and the part size it reported, both set
+    # by `BlobLibrary` once it has connected to the data plane.
+    _data_plane: BlobDataPlaneStub
+    _part_size: int
+
+    def authorizer(self):
```
rjhuijsman: please spell out the return type for _all_ methods we write

reboot/std/blob/v1/blob.py
```diff
@@ -0,0 +1,625 @@
...
+        # Every method is listed explicitly so none can be
+        # accidentally left ungated. `Create`, `SetDownloaders`, and
+        # the internal workflow methods are app-internal only; the
+        # upload-side methods and `Remove` require the caller to be the
+        # blob's uploader, unless `uploader_id` is empty; `DownloadUrl`
+        # is gated on the `downloader_ids` allow-list, or open when none
+        # was recorded; `Info` is visible to anyone who may upload or
+        # download the blob (either predicate suffices).
```
rjhuijsman: Everything after "ungated" seems duplicative of the code and likely to go stale; drop it. If you want to add _why_-explanations for any authorizer choice do so right before that line of the authorizer.

reboot/std/blob/v1/blob.py
```diff
@@ -0,0 +1,625 @@
...
+            _CONFIGURATION_RETRY_SECONDS /
+            _CONFIGURATION_RETRY_INTERVAL_SECONDS
+        )
+        last_error: Optional[AioRpcError] = None
+        for _ in range(attempts):
+            try:
+                return await stub.Configuration(ConfigurationRequest())
+            except AioRpcError as error:
+                last_error = error
+                await asyncio.sleep(_CONFIGURATION_RETRY_INTERVAL_SECONDS)
```
rjhuijsman: Use our `backoff` library instead

documentation/docs/learn_more/call/from_react.mdx
```diff
@@ -100,14 +100,33 @@ api = API(
...
+  // uploading; render their progress reactively via `Blob.Info`.
+  repeated string attachment_blob_ids = 2;
+}
+
 message ChatRoom {
   option (rbt.v1alpha1.state) = {
   };
-  repeated string messages = 1;
+  reserved 1;
+  repeated Message messages = 2;
```
rjhuijsman: Don't keep backwards compatibility for examples; they should always show the "if you're writing this from scratch, do it this way" example, and as you can see here they're used in the docs.

reboot/examples/chat-room/backend/src/chat_room_servicer.py
```diff
@@ -45,11 +45,14 @@ async def send(
                 raise ChatRoom.SendAborted(
                     AttachmentTooLarge(max_size_bytes=MAX_ATTACHMENT_BYTES)
                 )
+            # Only `size`: it and `max_size` are one `oneof`, so
+            # setting both would silently drop this one. An exact size
+            # is its own ceiling, and the check above already enforces
+            # the room's limit.
```
rjhuijsman: This comment isn't very interesting to readers of an example; it just says what not to do. Drop it.

rbt/std/blob/v1/BUILD.bazel
```diff
@@ -74,6 +75,26 @@ js_reboot_library(
     ],
 )
 
+# The browser client. `reboot/std/react/blob` uses it to drive the
+# upload protocol from a plain class, where React's rules put the
+# generated hooks out of reach.
```
rjhuijsman: Don't say where things will get used. These comments go stale. Reinforce your rule against this behavior.

rbt/std/blob/v1/BUILD.bazel
rjhuijsman, line 1: I think this file is in the wrong fixup; it should be targeting the frontend library's commit.

reboot/std/package.json
rjhuijsman, line 1: Ditto, this should be in a fixup targeting the frontend commit.

reboot/cli/commands/serve.py
```diff
@@ -381,8 +404,16 @@ async def serve_run(
...
+        try:
+            async with subprocesses.exec(*args, env=env) as process:
+                return await process.wait()
+        finally:
+            for data_plane_task in data_plane_tasks:
+                data_plane_task.cancel()
+                try:
+                    await data_plane_task
+                except asyncio.CancelledError:
+                    pass
```
benh: You should use @onelxj's new cancellation stuff! Although I ideally it isn't necessary because we can just have reboot.py be the one place that can start a filesystem data plane.

tests/reboot/std/blob/v1/blob_tests.py
```diff
@@ -0,0 +1,733 @@
...
+        """Fetches upload instructions, waiting for the `BeginUpload`
+        workflow to have provisioned the upload session."""
+        while True:
+            response = await blob.get_part_upload_instructions(
+                self.context,
+                part_numbers=part_numbers,
+            )
+            if response.ready:
+                return response
+            await asyncio.sleep(0.05)
```
benh: All of these helpers that poll, can we please use `.reactively()`?

documentation/docs/implement/application.mdx
```diff
@@ -13,19 +13,19 @@ Your entrypoint will construct an `Application` and then `run` it:
...
 <!-- MARKDOWN-AUTO-DOCS:START
-(CODE:src=../../../reboot/examples/chat-room/backend/src/main.py&lines=25-33) -->
+(CODE:src=../../../reboot/examples/chat-room/backend/src/main.py&lines=26-34) -->
 <!-- The below code snippet is automatically added from ../../../reboot/examples/chat-room/backend/src/main.py -->
 
 ```py
 async def main():
     await Application(
         servicers=[ChatRoomServicer],
+        # Message attachments are stored as blobs; `rbt dev run` and
```
benh, line 23: This seems a little dense for the docs but I guess we are trying to show a complete example?

rbt/std/blob/v1/blob.proto
```diff
@@ -0,0 +1,451 @@
...
+    uint64 size = 3;
+
+    // An upper bound, when the exact total is not known up front.
+    uint64 max_size = 4;
+  }
+
+  // The ID of the user that may upload into this blob. Empty means
+  // anyone who knows this blob's ID may upload; see the authorizer
+  // note in `reboot.std.blob.v1.blob`.
+  string uploader_id = 5;
```
benh: Why empty instead of explicitly optional?

rbt/std/blob/v1/blob.proto
```diff
@@ -0,0 +1,451 @@
...
+
+  // The IDs of the users who may download this blob. When unset
+  // (omitted at `Create`), anyone who knows the blob's ID may
+  // download. When set, only the listed users may (an empty list
+  // means no one but app-internal callers). The uploader is NOT
+  // implicitly a downloader. See the authorizer note in
+  // `reboot.std.blob.v1.blob`.
+  optional Downloaders downloader_ids = 10;
```
benh: Minor nit but downloader_ids.user_ids is a little confusing compared to downloaders.user_ids

rbt/std/blob/v1/blob.proto
```diff
@@ -0,0 +1,451 @@
...
+    uint64 size = 2;
+
+    // An upper bound in bytes, when the exact total is not known up
+    // front.
+    uint64 max_size = 3;
+  }
+
+  // The ID of the user that may upload into this blob. Leave empty to
+  // allow anyone who knows this blob's ID to upload.
+  string uploader_id = 4;
```
benh: optional?

rbt/std/blob/v1/blob.proto
```diff
@@ -0,0 +1,451 @@
...
+
+  // The IDs of the users who may download this blob. Omit the field
+  // entirely to let anyone who knows the blob's ID download it; set it
+  // (even to an empty list) to restrict downloads to the listed users.
+  // The uploader is NOT implicitly a downloader — omit them from the
+  // list if they shouldn't be able to download (e.g. a drop-box). Can
+  // be changed later with `SetDownloaders`.
+  optional Downloaders downloader_ids = 5;
```
benh: Rename to downloaders here and everywhere else?

rbt/std/blob/v1/blob.proto
```diff
@@ -0,0 +1,451 @@
...
+  Blob.Status status = 1;
+  string content_type = 2;
+
+  // Mirrors `Blob.size_limit`.
+  oneof size_limit {
+    uint64 size = 3;
+    uint64 max_size = 4;
+  }
+
+  string uploader_id = 5;
```
benh: Is this one required?

rbt/std/blob/v1/data_plane.proto
```diff
@@ -0,0 +1,195 @@
...
+}
+
+message DataPlaneCompleteUploadResponse {
+  // The committed object's composite ETag. Empty when `error` is set.
+  string etag = 1;
+
+  // A permanent-failure reason, if completion failed in a way the
+  // client can fix by re-uploading. When set, `etag` is empty and the
+  // control plane reverts the blob to UPLOADING with this message.
+  optional string error = 2;
```
benh: Seems like this should be a oneof.

reboot/aio/tests.py
```diff
@@ -163,6 +166,47 @@ def __init__(self) -> None:
...
+
+    async def stop(self) -> None:
+        try:
+            await super().stop()
+        finally:
+            if self._blob_data_plane is not None:
+                await self._blob_data_plane.stop()
+                # Only clear the env var if it still points at our data
+                # plane; another `Reboot` instance may have replaced it
+                # with its own in the meantime.
```
benh: Can you elaborate when and why this can occur?

reboot/cli/commands/dev.py
```diff
@@ -1638,6 +1643,58 @@ def crypto_root_keys() -> str:
...
+        subprocesses,
+        background_command_tasks,
+    )
+    # `start_filesystem_data_plane` reports the URL it bound by
+    # writing it into the environment it was given.
+    if ENVVAR_BLOB_DATA_PLANE_URL in data_plane_env:
+        env[ENVVAR_BLOB_DATA_PLANE_URL] = (
+            data_plane_env[ENVVAR_BLOB_DATA_PLANE_URL]
+        )
+
```
benh: Why do we need to possibly start a filesystem data plane here, or in test.py, or for serve? Can't we always do it in reboot.py if we are not running in the cloud or nobody has already specified  data plane URL that they already set up?

reboot/cli/commands/serve.py
```diff
@@ -325,6 +330,24 @@ async def serve_run(
...
+        # `--env`) and point the application at it. Bytes live under the
+        # state directory alongside the rest of the application's state.
+        # Done after `--env` is applied so an explicitly-configured data
+        # plane is honored rather than spawning a redundant local one.
+        blobs_directory = os.path.join(
+            env.get(ENVVAR_RBT_STATE_DIRECTORY, os.getcwd()),
+            BLOBS_SUBDIRECTORY,
+        )
+        data_plane_tasks: list[asyncio.Task] = []
+        await start_filesystem_data_plane(
```
benh: It was also weird to see this function but that was not what was used in test.py?

reboot/cli/common/blob_data_plane.py
```diff
@@ -0,0 +1,141 @@
...
+    try:
+        grpc_port, _ = await ready
+    except BaseException:
+        # Reap the never-ready server immediately: its task would
+        # otherwise run until the caller's cleanup — which under
+        # `rbt serve run` only happens once the application exits.
+        task.cancel()
+        try:
+            await task
+        except BaseException:
```
benh: This can still be a CancelledError, if that is not what you want use the new functions for waiting for something even if you get cancelled.

reboot/std/blob/v1/_store.py
```diff
@@ -0,0 +1,388 @@
...
+    def lock_for(self, encoded_blob_id: str) -> asyncio.Lock:
+        """Serializes one blob's completion against the part `PUT`s
+        that publish its bytes.
+
+        Both run in this one process -- the server hosts the gRPC
+        service and the byte endpoint together -- so an in-process
+        lock is enough to make completion see a fixed set of parts.
+        Held only across a `PUT`'s final rename, not across the upload
+        itself, so parts still upload concurrently."""
+        return self._locks.setdefault(encoded_blob_id, asyncio.Lock())
```
benh: Does this `dict` grow unbounded forever?

reboot/std/blob/v1/_store.py
```diff
@@ -0,0 +1,388 @@
...
+                    encoded,
+                    {
+                        "content_type": content_type,
+                        "committed": False,
+                        "upload_id": upload_id,
+                    },
+                )
+            return upload_id
+
+        return await asyncio.to_thread(sync)
```
benh: Why are you running this in its own thread? Is this just for `os.makedirs`? Can't we use an asyncio version like we do in the CLI IIRC? Or worst case just put `os.makedirs` in the thread, or just run it in an executor, why the thread?

reboot/std/blob/v1/_store.py
```diff
@@ -0,0 +1,388 @@
...
+        return os.path.join(self.blob_directory(encoded_blob_id), "meta.json")
+
+    def read_meta(self, encoded_blob_id: str) -> Optional[dict]:
+        try:
+            with open(self._meta_path(encoded_blob_id), "r") as f:
+                return json.load(f)
+        except FileNotFoundError:
+            return None
+
+    def _write_meta(self, encoded_blob_id: str, meta: dict) -> None:
```
benh: It would have been much easier to understand this if `meta` was a well defined type. Ideally a protobuf that you convert to and from JSON.

reboot/std/blob/v1/_store.py
```diff
@@ -0,0 +1,388 @@
...
+
+@dataclass(frozen=True)
+class UploadedPart:
+    """One part of an upload, as reported by the client."""
+    number: int
+    etag: str
+    size: int
+
+
+def _encode_blob_id(blob_id: str) -> str:
```
benh, line 103: Why not always just make blob_id safe?

rjhuijsman: Blob IDs can be picked by our customers, it's not necessarily the framework picking (it's a state ID) - so making `blob_id` safe would require that we reject certain blob IDs from being constructed. That's not a great development experience (e.g. an error you only discover at runtime, and may only discover in prod depending on how you construct the IDs) - and it's not a big problem for us to simply encode it.

tests/reboot/std/blob/v1/blob_tests.py
rjhuijsman: Done in 9f9f7674 -- both of them: the one waiting for `BeginUpload` to have provisioned the upload session, and the one waiting for a blob to reach a status.

No deadline on either, before or after. The test's own timeout already covers a condition that never becomes true, and a second lower-level one would only add a way to fail under load.

documentation/docs/implement/application.mdx
rjhuijsman: Fair, and fixed in dc836cb9. The example now says `# Message attachments are stored as blobs.` and nothing else, which puts the Python tab at the same density as the TypeScript one beside it.

What the comment was saying had somewhere better to be. The page never said what `libraries` *is*, though the snippet has always passed one -- so there is a short `libraries` section now, and the data plane and `REBOOT_BLOB_DATA_PLANE_URL` are described there in prose, rather than inside the code the reader is trying to read.

reboot/std/blob/v1/blob.py
```diff
@@ -0,0 +1,747 @@
...
+        # overwritten, never removed, so it would make `Commit`'s
+        # contiguity check fail forever.
+        if request.part_number < 1 or request.part_number > MAX_PARTS:
+            raise Blob.PartUploadedAborted(IncompleteParts())
+
+        # Validate the ETag as an MD5 hex digest, as the data-plane
+        # contract requires, so a client can't smuggle arbitrary
+        # content into the value a data plane later relies on to
+        # finalize the object.
+        if not _PART_ETAG_PATTERN.match(request.etag):
```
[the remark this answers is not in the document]
rjhuijsman: Good catch, and correct on both counts -- fixed in 67754a77.

`fullmatch` against an unanchored pattern now, having confirmed the hole: `re.compile(r'^[!#-~]{1,128}$').match("<md5hex>\n")` does return a match, so a trailing newline was getting through a check whose entire purpose is to refuse control characters. Test case added for exactly that value alongside the existing `"><injected/>` one.

The comment is reconciled the way you suggest: it claimed to validate "an MD5 hex digest, as the data-plane contract requires", which neither the pattern nor `data_plane.proto` agrees with. It now just points at `_PART_ETAG_PATTERN`, where the reason an ETag is opaque is already written down.

One correction to the report, for anyone reading it later: `_http.py` was already doing this correctly -- unanchored patterns with `fullmatch` -- so there was nothing to change there. (I briefly thought otherwise because I was looking at a stale copy of that file in another checkout.)

documentation/docs/call/from_react.mdx
```diff
@@ -402,15 +413,30 @@ Every call to a mutator returns both a `response` and an `aborted`; successful c
 [Learn more about Reboot errors and error types.](/errors)
 
 <!-- MARKDOWN-AUTO-DOCS:START
-(CODE:src=../../../reboot/examples/chat-room/frontend/web/src/App.tsx&&lines=37-41) -->
+(CODE:src=../../../reboot/examples/chat-room/frontend/web/src/App.tsx&&lines=181-200) -->
```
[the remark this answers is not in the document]
rjhuijsman: Fixed in b5bd7992 -- range widened to `lines=181-201` and the block regenerated with `make snippets`, so the committed snippet closes its `if (aborted !== undefined) {`.

Confirmed the line numbers before changing them: `return;` is line 200 and the closing `}` is 201, so the range was stopping exactly one line short. Thanks for noting that the directive change alone doesn't rewrite the body -- it doesn't, and the regenerated block is committed with it.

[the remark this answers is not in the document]
rjhuijsman: Same issue as the thread above; fixed in b5bd7992 by widening the range to `lines=181-201` and regenerating.

reboot/aio/applications.py
```diff
@@ -349,6 +349,18 @@ def __init__(
...
+                for servicer in library.legacy_grpc_servicers()
+            ]
+            if len(library_legacy_grpc_servicers) > 0:
+                if legacy_grpc_servicers is not None:
+                    legacy_grpc_servicers.extend(
+                        library_legacy_grpc_servicers
+                    )
+                else:
+                    legacy_grpc_servicers = library_legacy_grpc_servicers
+
```
rjhuijsman: Can we replace this whole thing with...
```suggestion
legacy_grpc_servicers = legacy_grpc_servicers or []
legacy_Grpc_servicers.extend(library_legacy_grpc_servers)
```
?

rbt/std/blob/v1/filesystem.proto
```diff
@@ -0,0 +1,204 @@
+// The filesystem blob data plane's own metadata.
+//
+// `data_plane.proto` defines the interface a data plane implements;
+// this is the state behind one particular implementation, the
+// filesystem data plane that Reboot applications get locally without
+// arranging storage for themselves. Bytes live on disk as part files.
```
rjhuijsman: Instead of "without ... themselves": "when no other data plane is provided".

reboot/aio/http.py
rjhuijsman, line 1: Are the diffs in this file intended to actually be part of 12c4ba656123e0161686fada15a5b5f45403f26b ?

reboot/std/blob/v1/_data_plane.py
```diff
@@ -2,22 +2,44 @@
...
+# The directory the data plane settled on, carried to the servers an
+# application spawns. Set by the application, not by anyone
+# configuring it: to say where blobs go, say where state goes.
+ENVVAR_BLOBS_DIRECTORY = "REBOOT_BLOBS_DIRECTORY"
 
 _SECURE_SCHEMES = ("https", "grpcs")
 
+# Where an unnamed run's blob bytes go, kept for the life of the
+# process so that every store built in it agrees on one directory.
+_temporary_directory: Optional[tempfile.TemporaryDirectory] = None
```
rjhuijsman: Even when the `.rbt` directory isn't going to store the state, doesn't Reboot store its state in a temporary directory already? Can we put the blobs in that directory too, so that we can compute the blob directory off the state directory on all servers, without having to set and pass `REBOOT_BLOBS_DIRECTORY`?

reboot/std/blob/v1/_data_plane.py
```diff
@@ -41,41 +63,169 @@ def channel_for_url(url: str) -> grpc.aio.Channel:
...
+
+
+@asynccontextmanager
+async def data_plane_stub(
+    context: Context | ExternalContext,
+) -> AsyncIterator[BlobDataPlaneStub | _NamedCaller]:
+    """A stub for whichever data plane this application uses.
+
+    `REBOOT_BLOB_DATA_PLANE_URL` names one that lives elsewhere -- on
+    Reboot Cloud it is the application's facilitator -- and is honored
```
rjhuijsman: Don't include Reboot Cloud implementation details in comments (or otherwise) in the public repo; those details will go stale.

reboot/std/blob/v1/blob.py
```diff
@@ -680,7 +717,10 @@ async def _configuration(
         deadline = time.monotonic() + _CONFIGURATION_RETRY_SECONDS
         while True:
             try:
-                return await stub.Configuration(ConfigurationRequest())
+                async with configured_data_plane_stub() as data_plane:
```
rjhuijsman: By calling `configured_data_plane_stub()` we're assuming that this it's NOT a local filesystem data plane, but one configured by URL? Why is that safe?

reboot/std/blob/v1/blob.py
```diff
@@ -643,35 +694,21 @@ async def pre_run(self, application: Application) -> None:
...
+            # Reverse-proxying a data plane's byte endpoints is no
+            # longer something this library does, so a data plane that
+            # needs it would mint URLs nothing serves. Say so now
+            # rather than at the first upload.
+            raise RuntimeError(
+                "This blob data plane asks for paths to be forwarded to "
+                "it, which is no longer supported; configure a data "
+                "plane that serves its own URLs via "
+                f"`{ENVVAR_BLOB_DATA_PLANE_URL}`."
             )
```
rjhuijsman: We never released this "forwarded_paths" feature; don't have a fallback for it, just remove the option entirely.

reboot/std/blob/v1/_data_plane.py
```diff
@@ -2,29 +2,46 @@
...
+
+
+def configured_data_plane_url() -> Optional[str]:
+    """The URL of the data plane this application has been pointed at
+    via `REBOOT_BLOB_DATA_PLANE_URL`, or `None` when it hosts one
+    itself."""
+    return os.environ.get(ENVVAR_BLOB_DATA_PLANE_URL) or None
 
 
 def channel_for_url(url: str) -> grpc.aio.Channel:
```
rjhuijsman, line 46: This is only called from `data_plane_stub_at`? If so, just inline it there.

reboot/std/blob/v1/_data_plane.py
```diff
@@ -41,41 +58,138 @@ def channel_for_url(url: str) -> grpc.aio.Channel:
...
-    url = os.environ.get(ENVVAR_BLOB_DATA_PLANE_URL)
-    if not url:
+
+    def __getattr__(self, name: str):
+
+        async def call(request):
+            return await getattr(self._stub,
+                                 name)(request, metadata=self._metadata)
+
+        return call
```
rjhuijsman: Would it not be neater to subclass `BlobDataPlaneStub` and simply override its `call` method?

reboot/std/blob/v1/_data_plane.py
```diff
@@ -41,41 +58,138 @@ def channel_for_url(url: str) -> grpc.aio.Channel:
...
+    if replica_config_json is None:
+        return 1
+    replica_config = placement_planner_pb2.ReplicaConfig()
+    # Tolerating unknown fields as the controller's own parse does, so
+    # that a newer controller adding one is not reported here as a
+    # parse error in place of the message this was going to give.
+    json_format.Parse(
+        replica_config_json, replica_config, ignore_unknown_fields=True
+    )
+    return len(replica_config.replicas)
```
rjhuijsman: Can we put this helper somewhere reusable, near other replica-related logic?

reboot/std/blob/v1/_data_plane.py
```diff
@@ -41,41 +58,138 @@ def channel_for_url(url: str) -> grpc.aio.Channel:
...
-            "start the filesystem blob data plane automatically), or "
-            "set the variable to your own data-plane service's URL."
+            "the blob data plane can only be reached from within an "
+            "application"
+        )
+    async with context.legacy_grpc_channel() as channel:
+        yield _NamedCaller(BlobDataPlaneStub(channel), application_id)
+
+
+def replicas() -> int:
```
rjhuijsman: I think `num_replicas()` is the clearer name

reboot/std/blob/v1/_filesystem_data_plane.py
```diff
@@ -0,0 +1,340 @@
...
+class FilesystemDataPlaneServicer(BlobDataPlaneServicer):
+    """Serves `BlobDataPlane` from the application whose blobs it
+    holds.
+
+    The store is set by `BlobLibrary` once it knows where this
+    application keeps them."""
+
+    _store: FilesystemBlobStore
+
+    async def _caller(self, context: LegacyGrpcContext):
```
rjhuijsman: Annotate the return type. All methods, where possible, should have their return type annotated - please note so in repo memory.

rjhuijsman: It's surprising that a method named `_caller` returns an `ExternalContext`. How about `_context_for_caller`? Or how about just a `_authorize_caller` that doesn't return anything, and when an external context is needed that's a separate call?

reboot/std/blob/v1/_filesystem_data_plane.py
```diff
@@ -0,0 +1,340 @@
...
+            response.upload_id,
+        )
+        return DataPlaneBeginUploadResponse(upload_id=response.upload_id)
+
+    async def GetPartUploadInstructions(
+        self,
+        request,
+        grpc_context: LegacyGrpcContext,
+    ):
+        context = await self._caller(grpc_context)
```
rjhuijsman: `context` is unused?

reboot/std/blob/v1/_filesystem_data_plane.py
```diff
@@ -0,0 +1,340 @@
...
+            etag = await self._complete(request, context)
+            return DataPlaneCompleteUploadResponse(etag=etag)
+        except BlobStoreError as error:
+            # A permanent failure: report it so the control plane can
+            # surface it and let the client re-upload. Transient
+            # failures raise other exceptions, which the control
+            # plane's workflow retries.
+            return DataPlaneCompleteUploadResponse(error=str(error))
+
+    async def _stored(self, blob_id: str, context):
```
rjhuijsman: Return type annotation here please; also lots of other places

reboot/std/blob/v1/_filesystem_data_plane.py
```diff
@@ -0,0 +1,340 @@
...
+            [(part.number, part.storage_id) for part in manifest],
+        )
+        return etag
+
+    async def GetDownloadUrl(
+        self,
+        request,
+        grpc_context: LegacyGrpcContext,
+    ):
+        context = await self._caller(grpc_context)
```
rjhuijsman: `context` is unused?

reboot/std/blob/v1/_filesystem_data_plane.py
```diff
@@ -0,0 +1,340 @@
...
+        # blob's own directory, so removing the directory removes any
+        # unfinished upload with it.
+        # Forgotten before the bytes go, so that nothing reads a
+        # manifest naming bytes that are already gone: between the two
+        # a download would answer `200` and then run out of file.
+        try:
+            await StoredBlob.ref(request.blob_id).always().forget(context)
+        except StoredBlob.ForgetAborted as aborted:
+            # Nothing was ever stored for this blob, so there is
+            # nothing to forget and deleting it has succeeded.
```
rjhuijsman: This comment is confusing; it seems to be describing the conclusion of getting a `ForgetAborted`, but really it's describing the conclusion of if the answer to the "if" statement below is "false". Consider restructuring this so this comment can sit _inside_ the branch of the "if" it's describing.

reboot/std/blob/v1/_http.py
```diff
@@ -84,114 +82,120 @@ def _unexpired_expiration(request: Request) -> Optional[int]:
...
+    """An app-internal context, which is what reaches `StoredBlob`.
+
+    Taken here rather than granted to the route, because a route is
+    granted one on the strength of its path: an application that
+    happens to serve its own handler at one of these paths -- under
+    another method, or through an earlier mount -- would be handed the
+    same privilege without having checked anything. Called only below
+    a verified signature, so what takes this has proven it holds a URL
+    this data plane minted."""
+    return request.state.reboot_app_internal_context(request)
```
rjhuijsman: Nothing is getting "verified" here? The comment also doesn't make much sense to me. Let's just drop this method and use its one-line implementation everywhere explicitly?

reboot/std/blob/v1/_http.py
```diff
@@ -84,114 +82,120 @@ def _unexpired_expiration(request: Request) -> Optional[int]:
...
+        )
+        if not published.published:
+            await store.discard_part(staged)
+            return Response(status_code=409, content="Blob already committed")
+
+        if published.HasField("superseded_storage_id"):
+            # This part had been uploaded before. Nothing is made of
+            # those bytes now, and the manifest that could still name
+            # them is refused at commit, so they go rather than
+            # accumulating a file per attempt.
```
rjhuijsman: Please reword "they go" - they go what?

reboot/std/blob/v1/_store.py
```diff
@@ -41,23 +45,30 @@
...
+# and part travel in the query rather than the path; all of them are
+# covered by the URL's signature either way.
+PART_PATH = "/__/reboot/blob/part"
+BLOB_PATH = "/__/reboot/blob"
 
 # HKDF `info` (domain separator) for the filesystem store's URL-signing
 # key.
 _SIGNING_INFO = b"reboot.std.blob.url-signing"
 
+_STREAM_CHUNK_SIZE = 1024 * 1024
```
rjhuijsman: What unit is this in?

reboot/std/blob/v1/_store.py
```diff
@@ -423,24 +299,180 @@ def download_url(
...
+                    size += len(chunk)
+                    await file.write(chunk)
+                await file.flush()
+                # `aiofiles` has no `fsync`; `fileno()` is proxied
+                # straight through, so the descriptor is the real one.
+                await asyncio.to_thread(os.fsync, file.fileno())
+        except BaseException:
+            # Never leave a partial file behind to be mistaken for a
+            # part.
+            await asyncio.to_thread(_unlink_if_present, temporary)
```
rjhuijsman: Does `aiofiles` not have an equivalent of `_unlink_if_present`?

reboot/std/blob/v1/_store.py
```diff
@@ -423,24 +299,180 @@ def download_url(
...
+            raise
+
+        path = self.part_path(
+            encoded_blob_id, upload_id, part_number, storage_id
+        )
+        # Renamed into place rather than written there, so a download
+        # never catches a part half-written. Safe to do before the
+        # manifest claims these bytes, because the name is theirs
+        # alone: at worst they are left unclaimed.
+        await asyncio.to_thread(_publish, temporary, path)
```
rjhuijsman: Does `aiofiles` not have an equivalent of `_publish`?

reboot/std/blob/v1/_store.py
```diff
@@ -423,24 +299,180 @@ def download_url(
...
+    async def discard_storage_id(
+        self,
+        encoded_blob_id: str,
+        upload_id: str,
+        part_number: int,
+        storage_id: str,
+    ) -> None:
+        """Drops one write of a part by name."""
+        await asyncio.to_thread(
+            _unlink_if_present,
```
rjhuijsman: same question about `aiofiles` here

reboot/std/blob/v1/_store.py
```diff
@@ -423,24 +299,180 @@ def download_url(
...
+        )
+
+    async def discard_part(self, staged: StagedPart) -> None:
+        """Drops a part's bytes, for one the object turned out not to
+        be made of.
+
+        Safe because the name belongs to this write alone: no manifest
+        can be pointing at it unless this write's own `PublishPart`
+        succeeded."""
+        await asyncio.to_thread(_unlink_if_present, staged.path)
```
rjhuijsman: Same question about `aiofiles` here

reboot/std/blob/v1/_store.py
```diff
@@ -423,24 +299,180 @@ def download_url(
...
+                    continue
+                if name.endswith(".partial"):
+                    # A write still in flight. Its own writer removes
+                    # it if it fails and renames it if it succeeds;
+                    # taking it here would turn that writer's refusal
+                    # into a failure to rename.
+                    continue
+                _unlink_if_present(os.path.join(directory, name))
+
+        await asyncio.to_thread(sync)
```
rjhuijsman: Could this not be done with `aiofiles` instead?

Same question goes for all uses of `asyncio.to_thread` for filesystem operations

reboot/aio/tests.py
```diff
@@ -163,6 +164,11 @@ def __init__(self) -> None:
         os.environ[ENVVAR_REBOOT_IN_TEST] = 'true'
         # The application under test, or `None` before one is started.
         self._application: Optional[Application] = None
+        # Whether `up()` has published this harness's state directory
+        # as `RBT_STATE_DIRECTORY`, and what that variable held before,
+        # for `stop()` to put back.
+        self._published_state_directory = False
+        self._state_directory_before_up: Optional[str] = None
```
rjhuijsman: This is plumbing to undo one of _many_ environment changes. How about we either...
A) don't bother, like we seem to do for all of the env stuff above.
B) create a full copy of the env before we start this test, and restore it after, so it works for everything

Take whichever one you feel is the least disruptive for what we need to achieve. If you choose option B, make it a separate commit.

reboot/std/blob/v1/_store.py
```diff
@@ -106,22 +107,24 @@ def _encode_blob_id(blob_id: str) -> str:
...
-def _unlink_if_present(path: str) -> None:
+async def _unlink_if_present(path: str) -> None:
     try:
-        os.unlink(path)
+        await aiofiles.os.unlink(path)
     except FileNotFoundError:
         pass
 
 
 def _fsync_directory(path: str) -> None:
```
rjhuijsman: Do we ever use `_fsync_directory` from a non-async context? If not, then just make it an `async def` that runs its own thread.

reboot/std/react/blob/index.tsx
```diff
@@ -0,0 +1,441 @@
...
+          this.context,
+          { partNumbers },
+          { signal: controller.signal }
+        );
+      for await (const response of responses) {
+        if (response.ready) {
+          return this.instructionsFrom(response);
+        }
+      }
+      options?.signal?.throwIfAborted();
```
benh, line 123: Why is this here? Please add a comment.

reboot/std/react/blob/index.tsx
```diff
@@ -0,0 +1,441 @@
...
+    // flips rather than asking again on a timer.
+    const first = await this.blob.getPartUploadInstructions(
+      this.context,
+      { partNumbers },
+      options
+    );
+    if (first.ready) {
+      return this.instructionsFrom(first);
+    }
+    options?.signal?.throwIfAborted();
```
benh, line 94: Why is this here? Please add a comment.

reboot/std/react/blob/index.tsx
```diff
@@ -0,0 +1,441 @@
...
+  async partUploadInstructions(
+    partNumbers: number[],
+    options?: { signal?: AbortSignal }
+  ): Promise<{ partSize: number; urls: Map<number, string> }> {
+    // Asked plainly first: a blob that is no longer uploading refuses
+    // with a declared error, which a plain call surfaces and a
+    // reactive read would retry forever. Only a blob whose upload
+    // session is still being provisioned is then watched: `ready` is
+    // false until the session has been provisioned, so watch until it
+    // flips rather than asking again on a timer.
```
benh: Reactive readers retrying forever if there is an error would be a giant bug that needs to be fixed!

reboot/std/react/blob/index.tsx
```diff
@@ -0,0 +1,441 @@
...
+      );
+    }
+    return { partSize: Number(response.partSize), urls };
+  }
+
+  /**
+   * `PUT`s one part's bytes to the data plane and reports it to the
+   * control plane. Idempotent per part number.
+   */
+  async putPart(
```
benh, line 146: Why do we need `putPart`? It seems like we're only ever calling `putPartToUrl`.

reboot/std/react/blob/index.tsx
```diff
@@ -0,0 +1,441 @@
...
+  /**
+   * Waits for the verdict on a commit under way: the blob's ETag, or
+   * the reason the commit failed.
+   */
+  private async commitVerdict(options?: {
+    signal?: AbortSignal;
+  }): Promise<UploadResult> {
+    // Subscribe rather than re-read on a timer: `Info` is a reader,
+    // so the update is pushed.
+    options?.signal?.throwIfAborted();
```
benh, line 191: Why is this here?

reboot/std/react/blob/index.tsx
```diff
@@ -0,0 +1,441 @@
...
+        if (
+          info.status === Blob_Status.REMOVING ||
+          info.status === Blob_Status.REMOVED
+        ) {
+          // Removal is not a verdict on the commit: there is no blob
+          // left to repair or to commit again.
+          throw new Error(`Blob ${this.options.blobId} has been removed`);
+        }
+      }
+      options?.signal?.throwIfAborted();
```
benh, line 223: Why is this here?

reboot/std/react/blob/index.tsx
```diff
@@ -0,0 +1,441 @@
...
+      if (!this.confirmed.has(partNumber)) {
+        pending.push(partNumber);
+      }
+    }
+
+    // One `partUploadInstructions` call per window rather than one per
+    // part, and no more URLs minted ahead of use than a window's worth:
+    // the URLs are short-lived, so fetching them all up front would see
+    // the later ones expire before their turn.
+    for (let index = 0; index < pending.length; index += UPLOAD_CONCURRENCY) {
```
benh: Lets use `@supercharge/promise-pool` here.

reboot/std/react/blob/put.ts
```diff
@@ -0,0 +1,156 @@
...
+  | { ok: false; retry: boolean; remint: boolean; reason: string };
+
+/**
+ * What `putPartWithRetries` reaches the world through. Both default to
+ * the real thing; a test hands in its own.
+ */
+export interface PutPartDependencies {
+  fetch?: typeof globalThis.fetch;
+  sleep?: (ms: number, signal?: AbortSignal) => Promise<void>;
+}
```
benh: Is this the best practice here? I would have assumed doing some kind of mocking is the better approach but maybe not?

reboot/std/react/blob/put.ts
```diff
@@ -0,0 +1,156 @@
...
+  options?: { signal?: AbortSignal } & PutPartDependencies
+): Promise<string> {
+  const fetchImpl = options?.fetch ?? globalThis.fetch;
+  const sleep = options?.sleep ?? delay;
+  for (let attempt = 1; ; attempt++) {
+    const outcome = await tryPutPart(url, bytes, options?.signal, fetchImpl);
+    // Compared rather than negated: this package compiles without
+    // `strict`, and only an equality check narrows a discriminant
+    // then.
+    if (outcome.ok === false) {
```
benh: Why are we not doing:
```suggestion
    if (!outcome.ok) {
```

reboot/std/react/blob/put.ts
```diff
@@ -0,0 +1,156 @@
...
+        throw new Error(
+          `Part ${partNumber} upload failed after ${attempt} ` +
+            `attempt${attempt === 1 ? "" : "s"}: ${outcome.reason}`
+        );
+      }
+      if (outcome.remint) {
+        url = await remint();
+      }
+      const backoff = PUT_FIRST_RETRY_DELAY_MS * 2 ** (attempt - 1);
+      await sleep(backoff * (0.5 + Math.random() / 2), options?.signal);
```
benh: Use `Backoff` from rbt/v1alpha1/index.ts!

reboot/std/blob/v1/_filesystem_data_plane.py
```diff
@@ -0,0 +1,88 @@
...
+import grpc
+from reboot.aio.caller_id import CallerID
+from reboot.aio.headers import CALLER_ID_HEADER
+from reboot.aio.interceptors import LegacyGrpcContext
+from reboot.aio.internals.contextvars import get_application_id
+from reboot.std.blob.v1._data_plane_servicer import BlobDataPlaneServicer
+from reboot.std.blob.v1._store import FilesystemBlobStore
+
+
+class FilesystemDataPlaneServicer(BlobDataPlaneServicer):
```
benh, line 38: Minor nit:
```suggestion
class FilesystemBlobDataPlaneServicer(BlobDataPlaneServicer):
```

reboot/std/blob/v1/_data_plane_servicer.py
```diff
@@ -0,0 +1,246 @@
...
+
+@dataclass(frozen=True)
+class UploadedPart:
+    """One part of an upload, as reported by the client."""
+    number: int
+    etag: str
+    size: int
+
+
+class BlobStore(Protocol):
```
benh, line 54: If this is meant to be public can we put this in `reboot/std/blob/v1/store.py` and also make this file public for access to `BlobDataPlaneServicer`.

reboot/std/blob/v1/_store.py
```diff
@@ -0,0 +1,746 @@
+"""The filesystem blob store: the open-source blob data plane's
```
benh, line 1: Lets rename this file to `_filesystem_store.py`.

reboot/aio/http.py
```diff
@@ -291,6 +299,15 @@ async def external_context_middleware(request: Request, call_next):
...
                 )
+            # Offered rather than applied, for a handler that can
+            # establish a caller's right itself and only then wants to
+            # act on the application's behalf. Reaching for this is a
+            # handler saying it has done that; the route-level
+            # `app_internal=True` above, which grants the same thing
+            # on the strength of a path alone, cannot make that check.
+            request.state.reboot_app_internal_context = (
+                app_internal_external_context_from_request
+            )
```
benh, line 310: I don't think we need this escape hatch anymore.

reboot/std/blob/v1/_data_plane.py
```diff
@@ -0,0 +1,178 @@
...
+        # cannot see the bytes this one stores. Refuse rather than
+        # serve downloads that succeed or 404 depending on which
+        # replica the request reached.
+        raise RuntimeError(
+            'Blob bytes are stored on the local disk of a single '
+            'replica, so they cannot be served by an application '
+            f'running across {count} replicas. Point '
+            f"'{ENVVAR_BLOB_DATA_PLANE_URL}' at a data plane all "
+            'replicas share.'
+        )
```
benh, line 169: Lets clean this up to make it clear that a filesystem directory serving blobs is only for the filesystem blob data plane which is only for `rbt dev` or `rbt serve` not run via `rbt cloud` so rather than have error messages that makes it seem like maybe the user could configure their replicas differently we just assert that this should never happen. In fact, it would be great if this was `assert` instead of `RuntimeError` to make it super clear that these are invariants we expect.

reboot/std/blob/v1/_data_plane.py
```diff
@@ -0,0 +1,178 @@
...
+    if application_id is None:
+        raise DataPlaneNotConfigured(
+            "the blob data plane can only be reached from within an "
+            "application"
+        )
+    async with context.legacy_grpc_channel() as channel:
+        yield _NamedCallerStub(channel, application_id)
+
+
+def blobs_directory() -> str:
```
benh, line 136: This function seems like it should be in one of the filesystem specific files not a the generic data plane file!

PR #110, opened 2026-07-24
Make clear TS is alpha

onelxj:
Do not confuse readers and give it clear that Python is better option right now.

### DIFF
--- a/documentation/docs/overview.mdx
+++ b/documentation/docs/overview.mdx
@@ -54,16 +54,23 @@ point:
   app."
 >
   <Card
-    title="Get Started (TypeScript)"
-    icon={<TSLogo />}
-    to="/full_stack_apps/typescript"
-  />
-  <Card
     title="Get Started (Python)"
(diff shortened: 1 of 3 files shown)


documentation/docs/overview.mdx
```diff
@@ -53,18 +53,25 @@ point:
...
+  <Card
+    title="Get Started (TypeScript)"
+    icon={<TSLogo />}
+    to="/full_stack_apps/typescript"
+  />
 </CardSection>
 
+<div style={{ fontSize: "0.9rem", marginTop: "-2rem" }}>
+  TypeScript backend support is in{" "}
+  <Link to="/full_stack_apps/typescript">alpha</Link>, and some
```
rileysdev: Why is it `alpha` that links to TypeScript? I would expect `TypeScript backend support` to be the link and alpha to be bold maybe.

rileysdev (approved):

PR #111, opened 2026-07-28
Rework the app-building skills from measured agent behaviour

onelxj:
I've run the reboot agent on the same prompt a couple of times; the very first run (current main) cost $39 and took 45 min to build. Now we've removed the plan stage and started writing files early which produces still high-quality output (at least on the example I've used). The comparison was done against supabase project ($10 and ~16 min), but supabase didn't end up writing a backend servicer, only sql migration file + web, so arguably we are now on the same stage but with a much bigger skill-set.

With the skill changes I've been able to build the app in ~16 min and ~$12, which is a huge improvement already.

image

### DIFF
--- a/reboot/plugin/skills/web-app/SKILL.md
+++ b/reboot/plugin/skills/web-app/SKILL.md
@@ -87,6 +87,17 @@ as MCP).
 > (anything that is not a Reboot-minted access JWT) falls
 > through to yours.
 
+The imports, so you don't have to go looking for them:
+
+```python
+from reboot.aio.applications import Application
+from reboot.aio.auth.oauth_providers import (
+    Development,
+    Google,                       # or GitHub, Auth0, …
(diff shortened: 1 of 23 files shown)


reboot/plugin/skills/web-app/SKILL.md
```diff
@@ -203,59 +188,68 @@ same recipe.
...
-  </RebootClientProvider>
-);
-```
+All of it — the provider (and the `url` it must be given), the
+generated hook surface, sign-in/sign-out, reading typed errors, and
+why a hook's `id` must be real on every render — is in
+[`references/react-client.md`](references/react-client.md). Read it
+at the frontend step; don't reconstruct it from memory here.
+
+## Which References to Read, and When
```
onelxj, line 197: I'll apply the same to chat-app as well.

benh (approved):

PR #116, opened 2026-07-30
Fix start hook for Codex

### LINKED ISSUE #115: SessionStart hook fails on Codex
Produces: 

Image

`session-start.sh` expects `CLAUDE_ENV_FILE` to be present, which isn't true with Codex.

onelxj:
Fixes #115

### DIFF
--- a/reboot/plugin/hooks-handlers/session-start.sh
+++ b/reboot/plugin/hooks-handlers/session-start.sh
@@ -12,11 +12,11 @@
 # plugin's install directory, set for hook commands by Claude Code.
 #
 # Codex discovers and runs the same `hooks.json`, so this handler runs
-# there too. Codex sets `$CLAUDE_PLUGIN_ROOT`, but has no per-command
-# env file and so sets no `$CLAUDE_ENV_FILE`; there the same PATH
-# prepend is wired by the `shell_environment_policy.set.PATH` entry
-# that `install.sh` merges into `~/.codex/config.toml`, and this
-# handler is a quiet no-op.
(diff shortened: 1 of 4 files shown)


reboot/plugin/hooks-handlers/session-start.sh
```diff
@@ -11,12 +11,28 @@
 # inject env vars from a SessionStart hook. `$CLAUDE_PLUGIN_ROOT` is the
 # plugin's install directory, set for hook commands by Claude Code.
 #
+# Codex discovers and runs the same `hooks.json`, so this handler runs
+# there too. Codex sets `$CLAUDE_PLUGIN_ROOT`, but has no per-command
```
rileysdev: Codex sets `$CLAUDE_PLUGIN_ROOT`? That's surprising.

rileysdev: This seems to be contradicted by the tests.

onelxj: Good catch — you're right, and the comment was wrong. `$CLAUDE_PLUGIN_ROOT` and `$CLAUDE_ENV_FILE` are set by Claude Code, not Codex, which is exactly why the Codex test case is `{}` (neither variable set). Fixed the header comment in 0dc2421 to say that neither variable is present under Codex, so the guard makes this a quiet no-op and the PATH prepend is wired through `~/.codex/config.toml` instead.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #119, opened 2026-07-30
Fix PreToolUse hook failing on Codex (allow decision rejected)

### LINKED ISSUE #118: PreTooUse hook fails on Codex
Produces:

Image

  Plugin’s PreToolUse hook returns:

  {
    "hookSpecificOutput": {
      "hookEventName": "PreToolUse",
      "permissionDecision": "allow"
    }
  }

  Codex doesn’t accept "allow" as a supported permissionDecision, so it rejects the hook response.

onelxj:
Fixes #118

### DIFF
--- a/reboot/plugin/hooks/auto-approve.sh
+++ b/reboot/plugin/hooks/auto-approve.sh
@@ -19,9 +19,10 @@
 #
 # Guards that hold for every case:
 #
-#   - $CLAUDE_PLUGIN_ROOT must be set. Claude Code sets it to the
-#     plugin's real install path when invoking its hooks; without
-#     it the `.../skills/` prefix check is meaningless.
+#   - $CLAUDE_PLUGIN_ROOT must be set. Both Claude Code and Codex
+#     set it to the plugin's real install path when invoking these
+#     hooks; without it the `.../skills/` prefix check is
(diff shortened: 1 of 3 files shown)


reboot/plugin/hooks/auto-approve.sh
```diff
@@ -43,7 +43,29 @@ fi
 
 input=$(cat)
 
+# Claude Code and Codex share this plugin. Codex sets CLAUDE_PLUGIN_ROOT
```
onelxj: Codex doesn't set that env var

onelxj: Good catch — thanks. Dropped the `PLUGIN_ROOT` env-var check entirely.

Instead the hook now positively detects Claude Code from the payload: Claude Code always includes a `permission_mode` field in PreToolUse input, while Codex (sandbox model, no permission modes) doesn't. It emits `allow` only when that field is present and stays silent otherwise — the Codex-safe default. Since it keys on a payload field rather than an env var, it holds even if Codex exports `CLAUDE_PLUGIN_ROOT` for compatibility. Fixed in 34b2807.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #122, opened 2026-08-03
Add GitHub token to the Claude Review run

onelxj:
Without that, the run fails, see the comment.

The CI uses the config from the `main` branch, so it is expected that this review run will fail for that PR.


rjhuijsman (approved):
🙏🏻

PR #125, opened 2026-08-11
Reboot DDD Dashboard

rileysdev:


rbt/dashboard/v1/dashboard.proto
```diff
@@ -0,0 +1,176 @@
...
+
+////////////////////////////////////////////////////////////////////////
+
+// The information the developer dashboard needs to display data
+// about the Reboot application being developed.
+
+message FieldInfo {
+  string name = 1;
+  string type = 2;
+}
```
benh: What do you mean by "field"? I see this used by `MethodInfo.arguments` and `MethodInfo.returns`. I don't track how this works because `FieldInfo` is just a `name` and a `type`, so if it is itself a record with fields, then I guess that is all stored in the `type`? That's no bueno because if you have a nested type of a nested type of a nested type ... you see where I'm going with this, it's all just in `type`. So what is the benefit of having the first layer of fields? Very weird.

My suggestion is to send a JSON schema for each argument and the return type. Pydantic should be able to just generate that for you, and with Protobuf I think you'll have to get the robot to write a protobuf descriptor to JSON schema translation. Maybe we already have something like that in the repo? Worth looking. Otherwise I think it's mechanical. Then you can use an existing TypeScript library to render that type.

What is the plan if one of the nested fields is itself a type? Do we have some other place where all of those types are and can be viewed? This applies to errors too, since they are otherwise totally opaque and we can't see what fields are in that type. I think we basically need some way of distinguishing data types from state types, that actually sounds pretty awesome. Not only will it help teach users that state and data are two different things, but it'll also be even less reasons to need to look at the code (or at least the API).

rbt/dashboard/v1/dashboard.proto
```diff
@@ -0,0 +1,176 @@
...
+
+message MethodInfo {
```
benh: Why the `Info` suffix? Seems superfluous, can we drop it everywhere? You don't call your fields `method_infos` you just call them `methods`.

rbt/dashboard/v1/dashboard.proto
```diff
@@ -0,0 +1,176 @@
...
+  };
+
+  // The state types the developer's API files declare, which exist
+  // before the application does.
+  repeated StateTypeInfo state_types = 1;
+
+  // Why the API files could not be read, if they could not be. A
+  // half-written file is the normal case while someone is typing, and
+  // saying so beats showing nothing.
+  string error = 2;
```
benh: Don't let your robot be lazy, make the type optional so it's explicit.
```suggestion
  optional string error = 2;
```

rbt/dashboard/v1/dashboard.proto
```diff
@@ -0,0 +1,176 @@
...
+}
+
+message APIGetRequest {}
+
+message APIGetResponse {
+  repeated StateTypeInfo state_types = 1;
+  string error = 2;
```
benh:
```suggestion
  optional string error = 2;
```

rbt/dashboard/v1/dashboard.proto
```diff
@@ -0,0 +1,176 @@
...
+}
+
+message APIUpdateRequest {
+  repeated StateTypeInfo state_types = 1;
+  string error = 2;
```
benh:
```suggestion
  optional string error = 2;
```
Here and everywhere else that is applicable, I won't add suggestions to all of them.

rbt/dashboard/v1/dashboard.proto
```diff
@@ -0,0 +1,176 @@
...
+  // Why the API files could not be read, if they could not be. A
+  // half-written file is the normal case while someone is typing, and
+  // saying so beats showing nothing.
+  string error = 2;
+}
+
+message APIGetRequest {}
+
+message APIGetResponse {
+  repeated StateTypeInfo state_types = 1;
```
benh: Bigger picture, I don't think we should put all of the state types on the main page, I think we should just have a home page that says something like "review state and data types by selecting them on the left, or review behaviors, or ..."

Or maybe the home page will be the "call graph" view. 

Either way I think it's overwhelming to see all of the text that we currently have, especially as your app scales up.

That probably means that we should have finer grained readers for getting only what we need for the current page that we're on (and take advantage of the local cache). I think we'd want to do this anyway as apps grow and we add lots more information (like the call graph), so that we're transmitting less and have a faster first paint.

I also think that once we have the call graph users will likely click into other state type or state methods or data types or behaviors to learn more and they'll want to hit the back button to go to the previous state/data/method/behavior, etc, and I think it'll seem less overwhelming if each state/data type has its own page. What do you think?

rbt/v1alpha1/options.proto
```diff
@@ -53,8 +53,9 @@ message McpMethodOptions {
   optional bool resource = 2;
   // Custom name for the tool/resource (defaults to method name).
   optional string name = 3;
-  // Description for the tool/resource.
-  optional string description = 4;
+  // Deprecated: write `MethodOptions.description` instead. Read as a
+  // fallback when it is the only description present.
+  optional string description = 4 [deprecated = true];
```
benh, line 58: Please add the skills and the skills upgrade for this so that we get any existing apps to move their descriptions!

reboot/cli/commands/dashboard.py
```diff
@@ -0,0 +1,245 @@
...
+    while True:
+        async with subprocesses.exec(
+            sys.executable,
+            '-m',
+            'reboot.dashboard.main',
+            env=env,
+            # The application's own startup output would drown the one
+            # line this command prints; anything it writes to stderr
+            # still reaches the terminal.
+            stdout=asyncio.subprocess.DEVNULL,
```
benh: Comment doesn't make any sense to me, do you mean the dashboard's startup output? Or the application that the dashboard is a companion to? Is the idea that we want users to be able to read the stderr output so that is why we are sending stdout to devnull? That seems like we should be changing the logging of the dashboard then so that most things are just not logged unless you changed the logging level. At the very least I'd clear up the comment.

reboot/cli/commands/dashboard.py
```diff
@@ -0,0 +1,245 @@
...
+    # dashboard a Node.js application or serve their frontend.
+    for name in (
+        ENVVAR_RBT_NODEJS,
+        ENVVAR_RBT_FRONTEND_HOST,
+        ENVVAR_RBT_FRONTEND_DIST_PATH,
+        ENVVAR_RBT_FRONTEND_ROOT_PATH,
+    ):
+        composed.pop(name, None)
+
+    composed[ENVVAR_RBT_DEV] = 'true'
```
benh: I'd add a comment that this is intentional and why we decided to run the dashboard in dev mode instead of serve mode, i.e., authn/authz was too hard. And if we really can't run in serve mode at the very least lets turn off effect validation!

benh (changes requested):
Let me know if you'd like to find some time to discuss these!

benh:
Some more food for thought for this as I work through method call analysis: can we just use the `.rbtrc` for determining the API directory? It seems strictly simpler to just re-use existing mechanisms vs adding new ones. Presumably this will simplify this PR too.

reboot/dashboard/api_watcher.py
```diff
@@ -0,0 +1,190 @@
...
+    async def update_if_changed(alias: str) -> None:
+        nonlocal updated
+
+        current = (descriptions.state_types(), descriptions.error())
+        if current != updated:
+            updated = current
+            # Every write from a workflow needs an identity, and this
+            # one writes once per file that changed, on every
+            # iteration.
+            await API.ref(API_ID).per_iteration(alias).Update(
```
benh: Can we make this an inline writer instead?

rbt/dashboard/v1/dashboard.proto
benh: Scratch this one, I see there are some name conflicts without the `Info` suffix.

benh:
> Some more food for thought for this as I work through method call analysis: can we just use the `.rbtrc` for determining the API directory? It seems strictly simpler to just re-use existing mechanisms vs adding new ones. Presumably this will simplify this PR too.

I actually did this in my follow up PR so nothing for you to do here.

rbt/dashboard/v1/dashboard.proto
rileysdev: Part of the way that it works is to show, on one page, the app getting assembled as it comes into being from the robot. So I think an overview page of some kind is still useful.

Now what to do as the app grows? I'd be surprised if you thought it _wasn't possible_ to display this much information usefully on one page. There are useful one page apps that have more information than an app is every likely to. I think we could do this and still make it useful by adding filters or search. 

I guess put another way, what would be the purpose of looking at just one state type on it's own page that couldn't be gotten by clicking the navigation fragment? If there is no difference, than I think it's more powerful on one page. 

I like the idea of a back button, we can use react router to achieve that with navigation fragments.

I think most people will think it's more cool than overwhelming. My preference is to leave it on one page but I'd be fine going to a higher level overview. I'd want to give some thought to the first few minutes experience so it still feels like something substantial is being built that you don't have to click around a bunch to see.

reboot/cli/commands/dashboard.py
rileysdev: I thought it wasn't possible to serve this in serve mode, but now I think it is. Updating.

reboot/dashboard/frontend/src/main.tsx
```diff
@@ -0,0 +1,1134 @@
...
+  const live = !isLoading;
+
+  // What the developer's API files declare. Those exist before the
+  // application is generated, built or started, which is why they are
+  // what this page shows.
+  // The description travels as a `google.protobuf.Value`: the types
+  // in it are Pydantic's own JSON Schema, which proto has no business
+  // restating.
+  const stateTypes: StateType[] = useMemo(
+    () => parseDescription(response?.stateTypes?.toJson()),
```
benh: This makes no sense. Why is `parseDescription` taking `stateTypes`? If "description" is the top-level API, then I'd expect to be storing a bunch of `Description` and having a field called `descriptions`.

But "description" is a horrible name so lets please not do that. I'm also not so sure that `StateType` is a great name here either because I have no idea what this "state type" entails. I'm assuming this is a Pydantic JSON schema, so at the very least I think having the word `schema` somewhere in there would be great. But more importantly what is it a schema of?

reboot/dashboard/api_reader.py
```diff
@@ -0,0 +1,225 @@
...
+
+    package = os.path.dirname(filename).replace(os.sep, '.')
+
+    described = []
+    for type_name, type_obj in api.get_types().items():
+        definitions, refs = _schemas_of(_models_of(type_obj))
+
+        state = refs[type_obj.state]['$ref'].replace(DEFS, '')
+
+        state_type: dict = {
```
benh: So this is your `state_type_schema`? Or this is your `state_type_info` which describes a state type and has some nested fields which are JSON schema shaped? I don't know JSON schema so I don't know if this dict represents something that is straight up JSON schema or your own home grown format. If it is home grown this is exactly what should be described in protobuf, leaving google.protobuf.Value to only JSON schemas.

reboot/dashboard/api_reader.py
```diff
@@ -0,0 +1,225 @@
...
+                        'name': name
+                    } for name in definitions if name != state
+                ],
+            'methods':
+                [
+                    _describe_method(method_name, spec, refs)
+                    for method_name, spec in type_obj.methods.items()
+                    if isinstance(spec, MethodModel)
+                ],
+            '$defs': definitions,
```
benh: Does this mean that if multiple state types are using the same data type you ship it to the front end twice? Do you then deduplicate in the front-end? Or I guess the definitions are passed down below this is just "what the state or its methods use" so the duplication would be multiple state types that all use the same data types.

reboot/dashboard/frontend/src/description.ts
```diff
@@ -0,0 +1,336 @@
...
+  mcp: z.boolean(),
+  errors: z.array(Ref),
+  request: Ref.optional(),
+  response: Ref.optional(),
+  description: z.string().optional(),
+});
+
+const DataTypeSchema = z.object({ id: z.string(), name: z.string() });
+
+const StateTypeSchema = z.object({
```
benh: Okay I think this answers my question. This looks 100% home grown which means it should be in protobuf, the only thing that should be a google.protobuf.Value should be the JSON schema objects themselves because those have no business being re-described as protobuf objects (I agree with the robot there) but these are _our_ types.

reboot/dashboard/frontend/src/description.ts
```diff
@@ -0,0 +1,336 @@
...
+export interface Field {
+  name: string;
+  type: string;
+  optional: boolean;
+  description?: string;
+  link?: string;
+}
+
+// One of the developer's types that is not a state type: a request, a
+// response, an error, or anything those hold.
```
benh: Or a field, as in a state type or data type might have a nested field which his of this type or a nested field which is a list of this type or a map of this type, etc.

reboot/dashboard/frontend/src/description.ts
```diff
@@ -0,0 +1,336 @@
...
+
+  const root: SchemaNode | undefined = tree.root.children[0];
+  return root === undefined ? [] : rowsOf(stateType, root);
+};
+
+// Every type the developer wrote that is not a state type, by id.
+//
+// The state types' own state models are left out: those are what the
+// state page is, and a state type is not data to be held.
+export const dataObjects = (description: Description): DataObject[] => {
```
benh: It seems like a smell that we have to do all of this work when it seems like we know this when we are doing the API analysis in api_reader the first time. This seems like we have the wrong `StateTypeSchema` shape.

reboot/dashboard/frontend/src/description.ts
```diff
@@ -0,0 +1,336 @@
...
+import { z } from "zod";
+
+// JSON Schema is recursive and open-ended, so it is carried through
+// rather than described: what the page needs from it is resolved by
+// `SchemaTree` below.
+const JsonSchema = z.record(z.string(), z.unknown());
+
+const Ref = z.object({ $ref: z.string() });
+
+const MethodSchema = z.object({
```
benh: I think the robot tricked you. This doesn't seem like well defined JSON schema either, this seems like your own home grown format. That should be defined in our top-level API. It does seem like there is some JSON schema somewhere in here though, or at least JSON schema _refs_.

And maybe that is why the robot did it this way? Because it wanted an object with a `$ref` named field? So maybe this is because you somehow can pass this object into some existing JSON schema libraries and it can "parse" it? But I'm dubious, I think the robot has made your protobuf types be basically useless (`repeated google.protobuf.Value state_types`) which makes your python backend code be incredibly difficult to read and 100% not type checked because you've got this front-end only API with no backwards compatibility stuffed in this file.

benh (changes requested):
Let's discuss.

reboot/dashboard/api_reader.py
rileysdev: Yes, we deduplicate on the frontend. I think this is a simpler design than using references and only shipping them once.

benh (approved):

PR #127, opened 2026-08-14
Fix uncertain idempotency on transaction should retry without backoff

benh:
Bugs reported from @mtrifiro.

See each commit.


reboot/aio/contexts.py
```diff
@@ -856,9 +856,18 @@ def use(self) -> Iterator[None]:
     # opt out.
     outstanding_rpcs: int
 
-    # Whether or not the transaction enclosing this context should
-    # abort.
-    transaction_unrecoverable_abort: bool
+    # The abort that made the transaction enclosing this context
+    # unrecoverable, or `None` if it does not have to abort. Truthy
+    # exactly when the transaction must abort, so it also reads as the
+    # "should abort" flag it replaced.
```
onelxj: Comment which references the old behavior.
We can simply remove the following part.
```
so it also reads as the
# "should abort" flag it replaced.
```

reboot/std/collections/ordered_map/v1/ordered_map.py
```diff
@@ -99,6 +99,10 @@ def _create(
         state: Node.State,
         request: NodeCreateRequest,
     ) -> None:
+        # A node of degree < 2 can not be split, which `_insert_leaf`
+        # and `_insert_inner` would spin forever trying to do.
+        assert request.degree >= 2, f"Invalid `degree` {request.degree}"
```
onelxj: Let's use the helper function you've defined further down instead?

reboot/std/collections/ordered_map/v1/ordered_map.py
```diff
@@ -860,15 +879,13 @@ async def Create(
                 )
             return OrderedMapCreateResponse()
 
-        state.degree = (
-            request.degree if request.HasField("degree") else DEFAULT_DEGREE
-        )
+        state.degree = degree if degree is not None else DEFAULT_DEGREE
```
onelxj: Can we just spell it as `degree or DEFAULT_DEGREE`?

reboot/std/collections/ordered_map/v1/ordered_map.py
```diff
@@ -911,18 +928,18 @@ async def Insert(
...
+        message = self._check_degree(degree)
+        if message is not None:
+            raise OrderedMap.InsertAborted(
+                InvalidArgument(),
+                message=message,
             )
+
+        # Construct the map if not yet constructed.
+        if context.constructor:
+            state.degree = degree if degree is not None else DEFAULT_DEGREE
```
onelxj: The same

onelxj (approved):

reboot/std/collections/ordered_map/v1/ordered_map.py
```diff
@@ -838,19 +854,22 @@ async def Create(
...
-        if context.constructor:
-            if request.HasField("degree") and request.degree < 2:
-                raise OrderedMap.CreateAborted(
-                    InvalidArgument(),
-                    message="`degree` must be >= 2",
-                )
-        else:
+        degree = request.degree if request.HasField("degree") else None
+
+        message = self._check_degree(degree)
```
onelxj: We already have `_check_construction_options`, why don't we put `_check_degree` inside that function so all validation will be in one place?

PR #129, opened 2026-08-14
Transactions fixes

onelxj:
2 fixes were found during Matt's debugging and my own benchmarks. Check each commit individually.

The test introduced in the first commit was able to reproduce the `CHECK(inserted)` issue on the CI, but it is a pretty complicated race between transactions start, lookup and create which is pretty tough to repro deterministically, so I left the original test which should cover both issues more or less deterministically.


rbt/v1alpha1/nodejs.proto
```diff
@@ -14,6 +14,11 @@ message Context {
   bool app_internal = 6;
   optional bytes auth = 7;
   optional string workflow_id = 8;
+  // The outermost transaction this call belongs to, if any. Several
+  // transactions may be joined to one state at the same time, so
+  // anything caching per-state data for the duration of a call must
+  // key on this too.
```
benh: The phrase "may be joined to one state at the same time" I don't think is as clear as we should make it, lets be very explicit with the exact same terminology we already use. How about: "Multiple transactions can lock the same state in shared mode at the same time so we need the root (outermost) transaction ID to distinguish a call (and we need to distinguish calls when caching per-state data)." Something like that.

reboot/templates/reboot.py.j2
```diff
@@ -143,14 +143,22 @@ ongoing_transaction_states: dict[str, IMPORT_reboot_api.Model] = {}
...
 # Helper to get the `ongoing_transaction_states` dictionary key.
-# The key should contain both the state ID and the state type name
-# to avoid conflicts when multiple states share the same ID.
+# The key contains the state type name and the state ID to avoid
+# conflicts when multiple states share the same ID, and the root
+# transaction ID because several transactions may be joined to one
+# state at the same time (they join it "shared", and only upgrade to
+# "exclusive" once they write). Without the transaction ID they would
+# share a single entry, and whichever registered last would receive
+# the others' writer updates, silently losing their effects.
```
benh: I don't understand how any effects are lost? Please elaborate this comment (and the comment in `reboot/templates/reboot.ts.j2`).

I don't understand because my expectation is that if some transaction 1 (T1) acquires shared then T2 acquires shared then T1 must wait to upgrade until T2 has released its shared hold. Moreover, T2 will never also be able to upgrade because we only support one upgrader.

Is something else happening like two upgraders are proceeding? Or we are upgrading _before_ waiting for shared holders to release? Because those would be major bugs.

The only other bug I could imagine here is a shared hold is getting acquired _after_ an exclusive hold which is overwriting the exclusive's writes? But that would also be a major bug.

tests/reboot/pydantic/concurrent_transactions_same_state/test.py
```diff
@@ -0,0 +1,213 @@
...
+class ConcurrentTransactionsOnOneStateTest(unittest.IsolatedAsyncioTestCase):
+
+    async def asyncSetUp(self) -> None:
+        self.rbt = Reboot()
+        await self.rbt.start()
+        await self.rbt.up(
+            Application(servicers=[CounterServicer]),
+            servers=1,
+            # Re-running every writer to validate that its effects are
+            # deterministic would obscure what these tests are about.
```
benh: Why does it obscure them? As in, are you counting things and the double counting is obscure?

benh (approved):

PR #134, opened 2026-08-18
Fix two transaction hangs, and other CI flakes

rjhuijsman:
CI is currently ~unpassable because of CI flakes. Most notably `//tests/reboot/pydantic/concurrent_transactions_same_state:test_py` is ~80% flaky. An agent identified two hangs in the transaction code from those flakes. These each have their own commit to fix them, please see the individual commit descriptions for details.

Additionally, we address timeouts and flakes in MacOS CI that are due to overload by reducing its workload, removing tests that don't add meaningful coverage.

Note that while the first commit's case is proven with unit tests, the second is hard to repro so needs some human reasoning to validate.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -351,10 +351,19 @@ jobs:
         QUERY='tests(//tests/...)'
 
         # 2) Exclude these tags by default for MacOS.
+        #
+        # A test carries `platform-independent` when it exercises no
+        # behavior that could differ between Linux and MacOS, or when
+        # whatever platform-dependent behavior it does exercise is also
+        # exercised by a `medium`-or-smaller test that still runs here.
+        # Such a test spends time on our slowest and most expensive
+        # runners for signal that the Linux jobs — which keep running
+        # every test — already give us.
         QUERY="$QUERY \
         except attr(\"tags\",\"requires-docker\",//tests/...) \
         except attr(\"tags\",\"macos_not_supported\",//tests/...) \
         except attr(\"tags\",\"requires-linux-x86\",//tests/...) \
+        except attr(\"tags\",\"platform-independent\",//tests/...) \
         except attr(\"tags\",\"manual\",//tests/...)"
 
         # 3) If this is a PR labelled "reboot-release", also drop flaky
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -352,18 +352,18 @@ jobs:
 
         # 2) Exclude these tags by default for MacOS.
         #
-        # A test carries `platform-independent` when it exercises no
-        # behavior that could differ between Linux and MacOS, or when
-        # whatever platform-dependent behavior it does exercise is also
-        # exercised by a `medium`-or-smaller test that still runs here.
-        # Such a test spends time on our slowest and most expensive
-        # runners for signal that the Linux jobs — which keep running
-        # every test — already give us.
+        # `macos-nightly-only` marks the tests whose MacOS results we
+        # are content to have once a day instead of on every pull
+        # request. They are among our slowest, and MacOS is our
+        # slowest and most expensive runner, so here — on the critical
+        # path of every merge — they cost more than they are worth.
+        # They keep running on Linux in this same run, and on MacOS in
+        # `.github/workflows/macos_nightly.yml`.
         QUERY="$QUERY \
         except attr(\"tags\",\"requires-docker\",//tests/...) \
         except attr(\"tags\",\"macos_not_supported\",//tests/...) \
         except attr(\"tags\",\"requires-linux-x86\",//tests/...) \
-        except attr(\"tags\",\"platform-independent\",//tests/...) \
+        except attr(\"tags\",\"macos-nightly-only\",//tests/...) \
         except attr(\"tags\",\"manual\",//tests/...)"
 
         # 3) If this is a PR labelled "reboot-release", also drop flaky
--- a/reboot/aio/state_managers.py
+++ b/reboot/aio/state_managers.py
@@ -1547,11 +1547,11 @@ class Lock:
       transaction. Passing `deadline=None` means wait forever or until
       the task is cancelled.
 
-    - A waiter that has given up, i.e., whose `deadline` elapsed or
-      whose task was cancelled, is dropped rather than granted (see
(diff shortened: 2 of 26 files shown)


tests/reboot/examples/agent-wiki/BUILD.bazel
```diff
@@ -51,7 +51,16 @@ sh_test_in_working_directory(
     # In all our example tests, we initiate LocalEnvoy on the
     # default secure port 9991. We require 'exclusive' to not run
     # multiple tests, that use the same port, concurrently.
-    tags = ["exclusive"],
+    #
+    # The `uv sync`, `rbt generate`, `pytest` and `rbt dev run
+    # --terminate-after-health-check` steps are the same ones the
+    # `medium` `//tests/reboot/examples/chat-room:test` runs on
+    # MacOS; only the application code in between differs, so this
+    # is `platform-independent`.
```
rjhuijsman: Per the comment above, a test running `uv sync` can't be "medium"; we may need to bump the size of `chat-room:test`?

This `agent-wiki` test may still be left marked "platform-independent" because it doesn't add coverage relative to `chat-room:test` though; we may need to reword the reasoning to explain so.

rjhuijsman:
@benh can you bump this to the top of your queue? These flakes make CI unpassable for all other PRs, and you're clearly the appropriate reviewer for transaction logic changes.

.github/workflows/build_and_test.yml
```diff
@@ -351,10 +351,19 @@ jobs:
...
 
         # 2) Exclude these tags by default for MacOS.
+        #
+        # A test carries `platform-independent` when it exercises no
+        # behavior that could differ between Linux and MacOS, or when
+        # whatever platform-dependent behavior it does exercise is also
+        # exercised by a `medium`-or-smaller test that still runs here.
+        # Such a test spends time on our slowest and most expensive
+        # runners for signal that the Linux jobs — which keep running
+        # every test — already give us.
```
benh: This doesn't give me warm fuzzy feelings. In my experience it is crazy difficult to be able to objectively know whether or not you are doing something that is platform independent or not. In Reboot alone we have had differences we've had to work out for example with macOS file name case-insensitiveness and file watching and of course the big one multiprocessing where @onelxj had to navigate the different approaches across Linux and macOS. It seems pretty futile to expect a coding agent to be able to correctly make this distinction and a human to be able to agree, and I don't love the idea of kidding ourselves for something like testing that this is objective when it is totally subjective.

I'd like to suggest another approach: we bring back CI on `main` but only run it nightly (if there's new code). Can we get reports emailed to us of the nightly runs?  Then we can label jobs that we only run on macOS nightly vs needing to decide whether or not we think the test only touches platform-independent machinery.

reboot/aio/state_managers.py
```diff
@@ -5394,13 +5394,43 @@ async def _transaction_coordinator_abort(
...
+        may be in the database; and (3) best-effort sends `Abort` RPCs
+        to all participants.
         """
+        # To indicate that the transaction has aborted we resolve the
+        # future to `None`, so a participant `Watch` that already passed
+        # the "is it still in the map" check and is awaiting the future
+        # observes the abort rather than finding itself still listed to
+        # commit. Both statements are synchronous and come before
+        # anything that can be interrupted, so from here on every
+        # watcher has an answer no matter where a cancellation lands.
```
benh:
```suggestion
        # commit.
        #
        # IMPORTANT: we need to resolve the future even if we fail to
        # call the database or fail to call the participants, or get
        # cancelled while doing either of those things! Thus we resolve
        # the future _before_ doing anything else. 
```

reboot/aio/state_managers.py
```diff
@@ -5394,13 +5394,43 @@ async def _transaction_coordinator_abort(
...
+        #
+        # Answering before the database cleanup below lands is safe,
```
benh:
```suggestion
       # Resolving the future _before_ the database cleanup is safe,
```

reboot/aio/state_managers.py
```diff
@@ -1783,12 +1815,18 @@ async def _wait(
                 waiter.future.done() and not waiter.future.cancelled() and
                 waiter.future.exception() is None
             ):
```
benh: If I'm reading the code correctly it seems like we should be able to fix the issue here and not add the complexity of having to reason about skipping an "abandoned" waiter scattered across multiple functions. I don't think that waiting for a cancelled or timed out waiter to clean it self up will be a performance problem.

Looking at this code it seems like we could change this to something where we catch just a `BaseException` and then handle the case where we may have already been granted and need to release and then if the exception instance is a `asyncio.TimeoutError` then we `raise SystemAborted(Unavailable(), ...)` otherwise we just `raise`.

benh (changes requested):

.github/workflows/build_and_test.yml
rjhuijsman: Done. We now tag tests as `macos-nightly-only` instead of `platform-independent`, and a new nightly job runs only those tests and emails us a report. We don't run these on PRs so that we get our 51-minute speedup, but the nightly tests will flag flakes or surprising platform-dependent behavior.

reboot/aio/state_managers.py
rjhuijsman: I think `shield()` makes sense, so resolving it here.

rjhuijsman:
Good points, addressed!

I told the robot to create a new `macos-nightly-only` tag that we set instead of `platform-independent`, and to run (only) those tests every night. If you think we should run _all_ tests every night, let me know. As long as those are also run by PRs I don't think it adds much, so I went with only-the-ones-not-run-on-PRs for now.

tests/reboot/examples/agent-wiki/BUILD.bazel
```diff
@@ -30,11 +30,6 @@ osx_env.update({
 
 sh_test_in_working_directory(
     name = "test",
-    # MacOS runners need the full Bazel "large" budget because
-    # `uv sync` cold-fetches the locked dependency tree (and
-    # possibly a managed Python interpreter) into the test
-    # sandbox, which exceeds the default "medium" 300s timeout.
-    size = "large",
     data = test_packages + [
```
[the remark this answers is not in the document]
rjhuijsman: Me and my robot could not find any evidence that the comment about `uv sync` is true. Hence the reduction in target size.

benh (approved):
🙏

PR #145, opened 2026-08-31
Flow control for reactive readers

rjhuijsman:
Last year, when we played Copenhunt, we found https://github.com/reboot-dev/mono/issues/4754 - reactive readers would fall behind if servers produced updates faster than clients could consume them. It was a major issue for a real Reboot app. We fixed #4754 in #4770, but its fix surfaced a bug that prompted us to revert - the bug it surfaced was later found and [fixed](https://github.com/reboot-dev/reboot/commit/dc7e8932) by @onelxj.

This PR fixes [#4754](https://github.com/reboot-dev/mono/issues/4754) **once and for all** 😬😄. See individual commits for more details.


rbt/v1alpha1/react.proto
```diff
@@ -90,8 +116,17 @@ message WebSocketsConnectionResponse {}
...
   // Allows users to "watch" the response of a unary reader method for
   // changes. The current response of the method is sent when the
-  // stream is opened, then a new response is sent whenever the
-  // state changes in a way that causes the method's response to change.
+  // stream is opened. A client that sets
+  // `client_can_acknowledge_responses` must acknowledge every
+  // response before a next one is produced, and that next response
+  // reflects the _latest_ state, skipping any states that came and
+  // went while the client was busy. A client that does not set it
+  // gets every response as soon as it is produced.
```
rjhuijsman: Let's not promise that setting `client_can_acknowledge_responses` gives you _every_ repsonse - that's both not true (servers could still do flow control on nested calls between states) and not a guarantee we want to give the impression of ever having given. Just say that a client that does not set it gets responses at a rate determined by the server, regardless of whether the client can consume responses that fast.

reboot/aio/contexts.py
```diff
@@ -498,25 +501,59 @@ async def loop():
...
+                                    self._event.set()
+
+                                await self._used_response[task].wait()
+
+                            # Only now that the response has been used
+                            # do we ask for a next one, so that it
+                            # reflects the latest state rather than a
+                            # state that has already been superseded.
+                            # An older backend doesn't send an ID and
+                            # doesn't expect an acknowledgement.
```
rjhuijsman: The more important thing about this flow control, and the thing I'd like to document here, is that the client can't fall behind the server if it's slow (https://github.com/reboot-dev/mono/issues/4754). The "state that has already been superseded" is not itself the main problem we're solving.

reboot/aio/contexts.py
```diff
@@ -498,25 +501,59 @@ async def loop():
...
+                                    ):
+                                        raise
+                                    # We have reached a server that has
+                                    # no `AcknowledgeQueryResponse`,
+                                    # which we can do mid-stream when a
+                                    # deployment is rolled back.
+                                    # Returning starts a fresh read;
+                                    # that server's responses carry no
+                                    # ID and expect no acknowledgement.
+                                    return
```
rjhuijsman: This is a good catch, but not relevant: Reboot doesn't allow rollbacks of changes that add methods, only roll-forwards. Let's not bother with this edge case.

reboot/aio/react.py
```diff
@@ -338,31 +342,167 @@ async def _websocket_query(
...
+                    await websocket.send(response.SerializeToString())
+                    continue
+
+                response.query_response_id = str(uuid.uuid4())
+
+                acknowledged = asyncio.Event()
+                acknowledgements[response.query_response_id] = acknowledged
+
+                try:
+                    await websocket.send(response.SerializeToString())
```
rjhuijsman: Looks like this `await websocket.send(response.SerializeToString())` is in fact unconditional; it appears identically in both branches. So let's do it before we branch?

reboot/aio/react.py
```diff
@@ -338,31 +342,167 @@ async def _websocket_query(
...
+                    # Wait for the client to tell us that it is ready
+                    # for a next response. This is what stops us from
+                    # producing responses faster than the client can
+                    # consume them; while we wait here the state may
+                    # change any number of times, and the next
+                    # response we produce will reflect the latest of
+                    # those states.
+                    await acknowledged.wait()
+                finally:
+                    acknowledgements.pop(response.query_response_id, None)
```
rjhuijsman: under what circumstance do we need the `None` fallback? Isn't it guaranteed we always have this entry?

reboot/aio/react.py
```diff
@@ -338,31 +342,167 @@ async def _websocket_query(
...
+            ) from None
+
+        if authoritative_server != middleware.server_id:
+            # This is NOT the correct server. Fail.
+            await grpc_context.abort(
+                grpc.StatusCode.UNAVAILABLE,
+                f"Server '{middleware.server_id}' is not "
+                "authoritative for this request; server "
+                f"'{authoritative_server}' is.",
+            )
```
rjhuijsman: In other places we `raise SystemAborted(Unavailable, ...)`, is that equivalent from a gRPC status code perspective? If so, let's standardize on one way of failing rather than having two.

reboot/aio/react.py
```diff
@@ -425,61 +565,51 @@ def done_callback(_) -> None:
...
+                    #       e.g. if we could measure the throughput to
+                    #       the client we could proactively send
+                    #       multiple responses without waiting for an
+                    #       ack for each one, which would reduce the
+                    #       user-visible latency of updates.
+                    await acknowledged.wait()
+                finally:
+                    self._query_response_acknowledgements.pop(
+                        response.query_response_id,
+                        None,
```
rjhuijsman: Same question here as on the other: when do we need this fallback?

reboot/templates/reboot.py.j2
```diff
@@ -5308,13 +5311,38 @@ class {{ client.proto.state_name }}:
...
+                                    except IMPORT_grpc.aio.AioRpcError as error:
+                                        if error.code() != IMPORT_grpc.StatusCode.UNIMPLEMENTED:
+                                            raise
+                                        # We have reached a server that has no
+                                        # `AcknowledgeQueryResponse`, which we
+                                        # can do mid-stream when a deployment is
+                                        # rolled back. Start a fresh read; that
+                                        # server's responses carry no ID and
+                                        # expect no acknowledgement.
+                                        break
```
rjhuijsman: Same comment here: this branch can't be reached in Reboot applications due to backwards compatibility guarantees, so let's remove it.

reboot/web/index.ts
```diff
@@ -418,6 +419,73 @@ export function reactively<
...
+          // when a deployment is rolled back or we are rerouted to an
+          // older server. Retrying can only fail again, so give up and
+          // let the caller reconnect; the fresh `Query` gets responses
+          // without an ID, which expect no acknowledgement.
+          throw new AcknowledgementsUnimplemented();
+        }
+        throw new Error(
+          `Acknowledging query response '${queryResponseId}' failed ` +
+            `with HTTP status ${response.status}`
+        );
```
rjhuijsman: Same here; assuming `headers` are the same as they were on a request whose response got an acknowledgement ID, the acknowledgement method will exist (backwards compatibility guarantees prevent rollbacks that delete methods), so we don't need to handle that edge case.

reboot/web/index.ts
```diff
@@ -430,21 +498,43 @@ export async function* reactiveReader({
...
+  }
+
+  if (url.protocol === "https:" && !websockets) {
+    const acknowledgeHeaders = new Headers(headers);
+    acknowledgeHeaders.set("Content-Type", "application/json");
+
+    const responses = grpcServerStream({
       endpoint: url.toString(),
       method: "POST",
       headers,
```
rjhuijsman, line 528: Interesting, I don't see a specification here of what state ID this query is talking to (in the form of a `x-reboot-state-ref` header); if that indeed doesn't get set then react queries get routed randomly by our Envoy? In that case we DO need to handle the case where one request gets an acknowledgement ID on the response but then the actual acknowledgement call gets a "not found" on the method - random routing means the two may talk to different instances, unlike for traffic that is targeted to the same state ref.

tests/reboot/reactivity_test.py
```diff
@@ -83,6 +177,149 @@ async def test_reactive_get_all_state(self) -> None:
...
+            name="Robot",
+            adjective="reactive",
+        )
+
+        with query_requests_without_acknowledgements() as acknowledged:
+            await self.start_accumulating_adjectives(greeter, context)
+            self.assertEqual(["reactive"], await self.get_adjectives(1))
+
+            await greeter.SetAdjective(context, adjective="realistic")
+
```
rjhuijsman, line 398: May be worth explaining in a comment here here that the test does NOT set `_can_accumulate_next_adjective` again, which means the server would block if it awaited acknowledgements - but we can accumulate, so the server isn't awaiting acknowledgements.

reboot/web/index.ts
```diff
@@ -381,6 +381,7 @@ export function reactively<
         const queryRequest = new react_pb.QueryRequest({
           method,
           request: request.toBinary(),
+          clientCanAcknowledgeResponses: true,
```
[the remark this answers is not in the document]
rjhuijsman: `reboot-dev-bot` is correct here, Copilot is incorrect.

reboot/templates/reboot_react.ts.j2
```diff
@@ -909,21 +910,44 @@ class {{ client.proto.state_name | to_camel }}Instance {
...
-                expecteds[0].idempotencyKey
-              )
-            ) {
-              await expecteds[0].observed(() => {
+            //
+            // A single query response can report the idempotency keys of
+            // several mutations at once, because the backend aggregates the
+            // keys of every state it skipped over on its way to the state this
+            // response reflects. Two mutations may also share an idempotency
+            // key. So observe _every_ mutation whose key this response
```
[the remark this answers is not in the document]
rjhuijsman: Agreed, we'll tackle this as a separate issue when this becomes something we see in practice. In the meantime, it's actually pretty OK for a Reboot container to OOM - it'll just restart.

rbt/v1alpha1/react.proto
```diff
@@ -101,6 +137,16 @@ service React {
     };
   }
 
+  // Acknowledges that a client has processed a response from `Query`
+  // and is ready for a next one. Must be called on the same server
+  // that produced the response.
+  rpc AcknowledgeQueryResponse(AcknowledgeQueryResponseRequest)
```
onelxj: The naming is weird here, how about `QueryAcknowledgement` and then `QueryAcknowledgementRequest`?

onelxj (approved):

rbt/v1alpha1/react.proto
rjhuijsman: I don't think those names would be correct - we're acknowledging the _response_ to a query (i.e. client makes a query, server responds, client acknowledges the response). The suggested names imply that we're acknowledging queries, not responses to queries.

I considered `Acknowledge`, but thought that was too vague ("acknowledge what?"). So I don't have better ideas...

rjhuijsman: I'm going to therefore leave this as-is, but happy to do a follow-up PR to fix this if you can come up with a better name!

onelxj: > I'm going to therefore leave this as-is, but happy to do a follow-up PR to fix this if you can come up with a better name!

How about `ClientIsReady` or something like that then and leave the mechanism be the implementation detail?

onelxj (commented):
Hey @rjhuijsman let's chat about that tomorrow before you merge it.

rbt/v1alpha1/react.proto
rjhuijsman: I like that way of thinking! In other words: request "I'm ready for more" rather than "I've seen the last". I've landed on `ContinueQuery` as a nice name for that; what do you think?

rbt/v1alpha1/react.proto
```diff
@@ -19,20 +19,26 @@ message QueryRequest {
...
-  // client; see `Query`. Only meaningful on transports where a client
-  // can send more than one message per query, i.e. websockets. When
-  // this is set the other fields are ignored.
-  string continue_query_response_id = 4;
-
-  // Whether this client asks for each next response once it has
-  // consumed the one before it. A client that does not set this gets
-  // responses at whatever rate the server produces them, whether or
-  // not it can consume them that fast.
+  reserved 4;  // Was `continue_query_response_id`.
```
rjhuijsman: We never submitted/released this field; you don't need to stay backwards compatible with it. Just remove/replace it.

rbt/v1alpha1/react.proto
```diff
@@ -56,19 +62,34 @@ message QueryResponse {
   // TODO(benh): send these as bytes?
   repeated string idempotency_keys = 2;
 
-  // Unique ID of this response, which a client names to continue the
-  // query after it. Only set when the client set
+  reserved 4;  // Was `query_response_id`.
```
rjhuijsman: Ditto; we never released this

reboot/aio/contexts.py
```diff
@@ -342,6 +342,57 @@ def from_grpc_metadata(cls, metadata: GrpcMetadata) -> 'Participants':
...
+    response it has handled; the server credits every response up to
+    that one, so nothing is lost by not reporting each individually.
+    """
+
+    def __init__(self, stub, metadata: GrpcMetadata):
+        self._stub = stub
+        self._metadata = metadata
+        self._report: Optional[asyncio.Future] = None
+
+    async def processed(self, query_response) -> None:
```
rjhuijsman: Please add types to _all_ method parameters; remember that in a repo-level rule.

reboot/aio/contexts.py
```diff
@@ -342,6 +342,57 @@ def from_grpc_metadata(cls, metadata: GrpcMetadata) -> 'Participants':
         return participants
 
 
+class QueryReports:
```
rjhuijsman: Where does the word "Report" come from? It's a new term that doesn't point us at its actual action, `ContinueQuery`, and isn't repeated elsewhere. Can we pick a more consistent name? Alternatively, do we rename `ContinueQuery`?

reboot/aio/react.py
```diff
@@ -26,10 +26,72 @@
...
+
+
+async def _next_response(
+    responses: AsyncIterator[react_pb2.QueryResponse],
+) -> Optional[react_pb2.QueryResponse]:
+    """Returns the next response, or `None` once there are no more."""
+    try:
+        return await responses.__anext__()
+    except StopAsyncIteration:
+        return None
```
rjhuijsman: Is this just `anext(responses, None)`? Could we use that instead?

rbt/v1alpha1/react.proto
rjhuijsman: Done — dropped the `reserved` and renumbered, so `client_continues_query` is now 4 and `continue_query_sequence_number` is 5. No gaps left in `QueryRequest`.

rbt/v1alpha1/react.proto
rjhuijsman: Done — same here: `query_id` is 4, `sequence_number` 5, and the two stall fields from the following commit moved down to 6 and 7.

reboot/aio/contexts.py
rjhuijsman: Done — `stub: react_pb2_grpc.ReactStub` and `query_response: react_pb2.QueryResponse`. Added `.claude/rules/python-annotate-every-parameter.md` as a standalone commit (739d9f7e), covering the why (an un-annotated parameter switches `mypy` off for the whole function body) and the exception for `*args`/`**kwargs` forwarded to an already-typed callee.

Worth considering separately: `disallow_incomplete_defs` in `mypy.ini` would enforce this mechanically rather than by convention. I have not touched it here since turning it on repo-wide is its own change with its own fallout.

reboot/aio/contexts.py
rjhuijsman: "Report" came from nowhere good — you are right that it invented a term. I went with matching `ContinueQuery` rather than renaming the RPC, since `ContinueQuery` is the name you and onelxj converged on and it reads well from the client's side.

So: `QueryReports` -> `QueryContinuations`, `processed()` -> `continue_past()`, the `_report` field -> `_continuation`, and `__reports__` -> `__continuations__` in the generated client. Swept the same term out of `index.ts` too (`report` -> `continuation`, and the comment prose) and out of `react.proto`, which had picked up "reports"/"reported" in about eight places.

One deliberate exception: the stall notice in `index.ts` is now `logStall`/`LOGGED_STALL_MILLISECONDS` rather than `reportStall`. That one tells the developer something via `console.info` and has nothing to do with continuing a query, so borrowing either vocabulary would have been worse than a plain `log`.

reboot/aio/react.py
rjhuijsman: Yes, it is exactly that — `anext(responses, None)`, builtin since 3.10, and we are on 3.10.12. Removed the helper.

For the record on why it existed: the `try`/`except StopAsyncIteration` form was there to satisfy `mypy`, which flagged `response` and `sequence_number` as `possibly-undefined` after the `except ... : break`. `anext(..., None)` sidesteps that rather than working around it, so it is better on both counts.

rjhuijsman:
@onelxj and @benh, please...
* Have a look at [my latest issue comment](https://github.com/reboot-dev/mono/issues/4754#issuecomment-5616492783) and let me know if you agree it's still a good idea to do application-level flow control.
* Have a look at the new commits here which make flow control use a window, as discussed, with continuous pulling of state updates (as requested by @benh to make @onelxj's upcoming PR easier) and logging in both the backend and the client if things fall more than 100ms behind.

reboot/aio/react.py
```diff
@@ -26,10 +27,89 @@
...
+        self._room = QUERY_RESPONSE_WINDOW
+        self._has_room = asyncio.Event()
+        self._has_room.set()
+        # Sequence number of the next response to send, and of the
+        # last one the client has reported processing. The client has
+        # reported nothing until it does, hence -1.
+        self.sequence_number = 0
+        self._processed = -1
+
+    def try_take(self) -> Optional[int]:
```
benh: We should be able to use an `asyncio.Semaphore` to simplify this logic.

benh (changes requested):
Lets chat, I'd like to simplify a few things here to account for the upcoming changes in reactive readers and caching.

reboot/aio/react.py
rjhuijsman: 👍🏻

reboot/aio/contexts.py
```diff
@@ -497,26 +558,39 @@ async def query():
...
+                                self._calls[task] = call
 
-                            if not have_first_response.is_set():
-                                have_first_response.set()
-                            else:
-                                self._event.set()
+                                self._responses[task] = asyncio.Future()
+                                self._responses[task].set_result(response)
 
-                            await self._used_response[task].wait()
```
benh, line 519: Okay great glad to see this gone, but just curious did you confirm with your agent that there is no reason why we were doing that in the first place? I don't think there is but it would be great to get its assessment.

rbt/v1alpha1/react.proto
```diff
@@ -40,8 +59,46 @@ message QueryResponse {
   //
   // TODO(benh): send these as bytes?
   repeated string idempotency_keys = 2;
+
+  // Unique ID of the query this response belongs to, which a client
+  // names when it continues that query. The same for every response
+  // of one `Query`. Only set when the client set
+  // `client_continues_query` on its `QueryRequest`; an empty ID means
+  // the client is not expected to continue anything.
+  string query_id = 4;
```
benh:
```suggestion
  optional string query_id = 4;
```

reboot/aio/contexts.py
```diff
@@ -342,6 +342,65 @@ def from_grpc_metadata(cls, metadata: GrpcMetadata) -> 'Participants':
...
+        self._continuation: Optional[asyncio.Future] = None
+
+    async def continue_past(
+        self,
+        query_response: react_pb2.QueryResponse,
+    ) -> None:
+        """Continues the query past `query_response`, if there is no
+        call already in flight. Raises whatever a previous call raised,
+        so that a failure to continue surfaces on the read."""
+        if query_response.query_id == '':
```
benh:
```suggestion
        if query_response.HasField('query_id'):
```

reboot/aio/contexts.py
```diff
@@ -342,6 +342,65 @@ def from_grpc_metadata(cls, metadata: GrpcMetadata) -> 'Participants':
...
+                self._stub.ContinueQuery(
+                    react_pb2.ContinueQueryRequest(
+                        query_id=query_response.query_id,
+                        sequence_number=query_response.sequence_number,
+                    ),
+                    # The same metadata ensures we're routed to the
+                    # same server.
+                    metadata=self._metadata,
+                )
+            )
```
benh, line 392: This looks like a bug, consider:

(1) First call to `continue_past` sets `self._continuation` then returns.
(2) Second call to `continue_past` where `self._continuation` is not yet done so it falls off the end of the function and doesn't do anything.

Maybe this never actually stalls in practice but in the degenerate case I do beleive this could cause an infinite hang.

How about a simple async function that we run that reads from a queue? Or if you want to only send the latest then it can just drain the queue after it gets awoken via `not queue.empty()` and `queue.pop_nowait()` and then send the last one? Or some other but still having a long-running asyncio task that is responsible for making the `ContinueQuery` call.

reboot/aio/contexts.py
```diff
@@ -342,6 +342,65 @@ def from_grpc_metadata(cls, metadata: GrpcMetadata) -> 'Participants':
...
+
+    async def stop(self) -> None:
+        if self._continuation is not None:
+            self._continuation.cancel()
+            try:
+                await self._continuation
+            except BaseException:
+                pass
```
benh: Lets use @onelxj's `wait_for_tasks` please!

reboot/web/index.ts
```diff
@@ -418,6 +419,86 @@ export function reactively<
   return [responses(), setRequest];
 }
 
+// How long a response must have waited for us before we say so in
+// the console. 100ms is about where a person stops experiencing an
+// update as immediate and starts perceiving lag, so a response that
+// waited longer than this is one whose delay our user could see; the
+// backend uses the same threshold for its own log.
+const LOGGED_STALL_MILLISECONDS = 100;
```
benh: Where did this number come from? Did you test what 100ms delays look like when using `Presence` and the mouse tracker?

I see this number, and even the size of the window, as things we need to get experience with.

For now I think we should always log it until we have that experience.

Honestly I think what I'd love to see the log message say something along the lines of "if dropping messages is not what you expected from this reactive read then you either need to improve the performance of your app so you can handle the incoming responses or if flow control is a feature for you then you can disable the console warning by passing `{ warnOnFlowControl: false }` to your `useReader()`"

Of course that is more work but then we can have our cake and eat it too, for folks that are trying to do something like `Presence` and mouse tracking they can make sure they improve their performance and for folks where the last response is just fine and flow-control is a feature then they can drop the log statements.

reboot/web/index.ts
```diff
@@ -418,6 +419,86 @@ export function reactively<
...
+
+// Tells the developer that this client fell far enough behind its
+// backend that the backend merged updates it would otherwise have
+// sent. The state we go on to render is still the latest one; what
+// was lost is the updates on the way there.
+function logStall(method: string, response: react_pb.QueryResponse) {
+  if (response.stallMilliseconds > LOGGED_STALL_MILLISECONDS) {
+    console.info(
```
benh:
```suggestion
    console.warn(
```

reboot/web/index.ts
```diff
@@ -430,21 +511,73 @@ export async function* reactiveReader({
...
-    const headers = new Headers();
 
-    if (request.bearerToken !== undefined) {
-      headers.set("Authorization", `Bearer ${request.bearerToken}`);
-    }
+  const headers = new Headers();
+
+  if (request.bearerToken !== undefined) {
+    headers.set("Authorization", `Bearer ${request.bearerToken}`);
+  }
```
benh, line 532: Why was this pulled out of the `if`? It makes it look like these headers are used in both the `if` and `else` branch but they aren't.

reboot/web/index.ts
```diff
@@ -430,21 +511,73 @@ export async function* reactiveReader({
...
+          sequenceNumber: response.sequenceNumber,
+          signal,
+        }).finally(() => {
+          settled = true;
+        });
+
+        // A failure surfaces on the `await` above, but a read that
+        // ends before we get there leaves it unobserved; observe it
+        // here too so that it can't become an unhandled rejection.
+        continuation.catch(() => {});
```
benh, line 591: I don't understand this ... this is silently swallowing any rejection which means it exactly becomes an "unhandled rejection". Am I missing something?

reboot/web/index.ts
```diff
@@ -430,21 +511,73 @@ export async function* reactiveReader({
...
+        settled = false;
+        continuation = continueQuery({
+          endpoint,
+          headers: continueHeaders,
+          queryId: response.queryId,
+          sequenceNumber: response.sequenceNumber,
+          signal,
+        }).finally(() => {
+          settled = true;
+        });
```
benh, line 586: This looks like it suffers from the same issue as the Python version of this code.

(1) A response comes in and we call `continueQuery`.
(2) The next response comes in and `continueQuery` is not done yet so it gets dropped.

reboot/aio/react.py
```diff
@@ -390,6 +559,149 @@ async def _query(
...
+                # that was already set, so clearing here would throw
+                # away the only notice we get, and the next iteration
+                # would wait on a task that has already ended.
+                if failure is None:
+                    have_next_response.clear()
+
+                stall_milliseconds = round(
+                    (time.monotonic() - stall_start) * 1000
+                )
+                if stall_milliseconds > REPORTABLE_STALL_MILLISECONDS:
```
benh: See other comment but tl;dr; lets always log it for now please.

reboot/aio/react.py
```diff
@@ -390,6 +559,149 @@ async def _query(
...
+                    logger.info(
```
benh:
```suggestion
                    logger.warning(
```

reboot/aio/react.py
```diff
@@ -390,6 +559,149 @@ async def _query(
...
+                response.skipped_updates = skipped_updates
+                response.stall_milliseconds = stall_milliseconds
+                yield response
+
+        finally:
+            accumulator.cancel()
+            try:
+                await accumulator
+            except CancelledError:
+                pass
```
benh: Use `wait_for_tasks`.

reboot/aio/react.py
```diff
@@ -26,10 +28,61 @@
...
+# changes sooner, because the server doesn't wait to hear about the
+# previous response before sending the next; it also means a client
+# that can't keep up is further behind, since it works through what is
+# already in flight before it sees current state. One is the most
+# conservative choice and behaves like a strict request-per-response
+# protocol.
+#
+# TODO: estimate this per client from observed throughput rather than
+#       fixing it for everyone.
+QUERY_RESPONSE_WINDOW = 10
```
benh, line 46: This seems low, will this stall perfectly healthy network links arbitrarily that could have kept up with each message?

benh (changes requested):

reboot/aio/contexts.py
rjhuijsman: There you go 😄

reboot/web/index.ts
rjhuijsman: 100ms was picked because it is, famously, the threshold below which interactions feel "instantaneous" according to the team that built GMail; see e.g. [this blog post](https://blog.superhuman.com/superhuman-is-built-for-speed/).

However, I see your point on that we may want to tune it ourselves, so I've asked my agent to add the flag you suggested and make the print unconditional.

rjhuijsman: One design decision I've made: the server-side warning gets controlled by a server-side setting (in the API file), the client-side setting you suggested only controls the client-side warning. The alternative (client-side warning controls both) has an awkward interaction with transitive reader calls (does the setting propagate through the whole stack? That's a lot of plumbing) plus "my client decides what my server warning-logs" also goes against my gut feeling.

rjhuijsman:
The robot is building https://github.com/reboot-dev/reboot/pull/173 to supersede this, doing only client-side coalescing for now so we don't have to think about complicated windowing logic to keep high-bandwidth clients maximally smooth.

PR #153, opened 2026-09-04
Fix several Maestro test flakes

rjhuijsman:
The `Test React Native mobile app (Maestro)` test was flaky; the robot traced it down to these issues:
1. The NPM registry was flaky, so some fetches stalled. The React Native tests run `rbt generate` which installs `npm` packages. The fix is to pre-fetch NPM packages into the cache once in the `Dockerfile`, then re-use that cache for all tests. Now there's only one fetch per test suite, so the odds of hitting an NPM registry issue are greatly reduced.
2. Under CPU contention, some deadlines were too tight. We loosen those deadlines somewhat.

While we're here, also tweak Claude's rules about how it writes PR descriptions/comments/... - that caused an issue in producing this PR.

### DIFF
--- a/.claude/rules/github-prose-no-hard-wrap.md
+++ b/.claude/rules/github-prose-no-hard-wrap.md
@@ -1,0 +1,24 @@
+# Don't hard-wrap prose posted to GitHub
+
+Anything that goes into GitHub's web UI — pull request descriptions,
+issue bodies, and every comment or review reply on either — gets one
+long line per paragraph, with no hard line breaks. GitHub soft-wraps
+it to the reader's viewport.
+
+Commit messages are the opposite: hard-wrap their bodies at ~72
+columns, because they are read in terminals through `git log`.
+
+**Why:** A hard-wrapped paragraph renders ragged on GitHub, because
+the browser wraps the already-broken lines a second time at a width
+unrelated to the one they were broken at, leaving short uneven lines
+that get worse on a narrow viewport. A terminal does no wrapping of
+its own, so a commit message needs exactly the breaks the web UI
+should never receive.
+
+**How to apply:** When writing a PR description, an issue, or a
+comment, keep each paragraph on a single line however long it runs;
+tables, lists, headings and fenced code blocks still break as their
+own syntax requires. Take particular care when a PR description is
+drafted alongside a commit message — the two are governed by opposite
+rules, and it is easy to carry the commit message's wrapping into the
+PR body.
--- a/Dockerfile
+++ b/Dockerfile
@@ -710,6 +710,22 @@ RUN if [ "${TARGETARCH}" = "amd64" ]; then \
     && bash /tmp/maestro-install.sh \
     && rm /tmp/maestro-install.sh; \
     fi
+
+# Warm the npm cache with the packages `rbt generate` installs into a
+# project's `.rbt` directory, and with their dependencies. This
+# version must match the one `_check_or_install_npm_packages` asks for
+# in `reboot/cli/commands/generate.py`. When a test runs `rbt dev run`
+# against a fresh state directory it installs these packages, and a
+# warm cache lets it do that without a registry round-trip. A real
+# resolution is what walks the dependency graph, so this installs the
+# three into a throwaway prefix, leaving the whole closure in the
+# cache; it also fails the build if the pin stops resolving.
+ARG BUFBUILD_VERSION=1.10.1
+RUN npm install --prefix /tmp/npm-cache-warm \
+    @bufbuild/protoplugin@${BUFBUILD_VERSION} \
+    @bufbuild/protoc-gen-es@${BUFBUILD_VERSION} \
+    @bufbuild/protobuf@${BUFBUILD_VERSION} \
+    && rm -rf /tmp/npm-cache-warm
 USER root
 
 ###############################################################################
--- a/reboot/examples/chat-room/frontend/mobile/.tests/maestro_test.sh
+++ b/reboot/examples/chat-room/frontend/mobile/.tests/maestro_test.sh
@@ -48,6 +48,23 @@ export ANDROID_HOME
 export ANDROID_SDK_ROOT="${ANDROID_HOME}"
 export PATH="${ANDROID_HOME}/cmdline-tools/latest/bin:${ANDROID_HOME}/platform-tools:${ANDROID_HOME}/emulator:${HOME}/.maestro/bin:${PATH}"
 
+# Keep a degraded npm registry from consuming this test's budget.
+# `prefer-offline` takes any package already in the cache without
+# revalidating it against the registry, which covers the `.rbt`
+# packages the devcontainer image pre-caches; only what the cache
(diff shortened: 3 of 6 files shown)


reboot/examples/chat-room/frontend/mobile/.tests/maestro_test.sh
```diff
@@ -48,6 +48,17 @@ export ANDROID_HOME
...
 export PATH="${ANDROID_HOME}/cmdline-tools/latest/bin:${ANDROID_HOME}/platform-tools:${ANDROID_HOME}/emulator:${HOME}/.maestro/bin:${PATH}"
 
+# npm's default `fetch-timeout` is five minutes, which a degraded
+# registry can burn twice over in one run: once for the `.rbt` install
+# that `rbt dev run` performs and once for the app's own dependencies.
+# That alone outlasts this test's budget, so fail a stalled request
+# quickly and retry it instead. `prefer-offline` then serves the
+# `.rbt` packages from the cache the devcontainer image warms, rather
+# than revalidating them against the registry.
+export npm_config_fetch_timeout=60000
```
rjhuijsman, line 64: Is this timeout in seconds? In that case the "fail quickly" doesn't seem true?

reboot/examples/chat-room/frontend/mobile/.tests/maestro_test.sh
```diff
@@ -48,6 +48,17 @@ export ANDROID_HOME
 export ANDROID_SDK_ROOT="${ANDROID_HOME}"
 export PATH="${ANDROID_HOME}/cmdline-tools/latest/bin:${ANDROID_HOME}/platform-tools:${ANDROID_HOME}/emulator:${HOME}/.maestro/bin:${PATH}"
 
+# npm's default `fetch-timeout` is five minutes, which a degraded
+# registry can burn twice over in one run: once for the `.rbt` install
+# that `rbt dev run` performs and once for the app's own dependencies.
+# That alone outlasts this test's budget, so fail a stalled request
+# quickly and retry it instead. `prefer-offline` then serves the
+# `.rbt` packages from the cache the devcontainer image warms, rather
+# than revalidating them against the registry.
```
rjhuijsman: I don't understand; `prefer-offline` only serves packages from the cache after retries? That doesn't sound very "prefer"; explain more clearly please.

Dockerfile
```diff
@@ -710,6 +710,20 @@ RUN if [ "${TARGETARCH}" = "amd64" ]; then \
...
     && rm /tmp/maestro-install.sh; \
     fi
+
+# Warm the npm cache with the packages `rbt generate` installs into a
+# project's `.rbt` directory. These versions must match the ones
+# `_check_or_install_npm_packages` asks for in
+# `reboot/cli/commands/generate.py`. The Maestro tests run `rbt dev
+# run` against a fresh state directory, so that install happens on
+# every run; a warm cache plus `prefer-offline` in the test harness
+# serves it without a registry round-trip. Only on amd64, matching the
```
rjhuijsman: Let's not make it specific for the Maestro tests, there's plenty of other tests that also install `npm` packages probably - or else there will be in the future. Let's just say "When tests run `rbt dev run` [...] it installs these packages without a registry round-trip" (and not be specific about which test)

Dockerfile
```diff
@@ -710,6 +710,20 @@ RUN if [ "${TARGETARCH}" = "amd64" ]; then \
...
+# SDK install above.
```
rjhuijsman: Why is the SDK install amd64 only? Our SDK works on arm64 too; can we remove this constraint?

reboot/examples/chat-room/frontend/mobile/.tests/maestro_test.sh
rjhuijsman: Milliseconds — npm's `fetch-timeout` is in ms, so `60000` is one minute, against npm's default of `300000` (five minutes). "Quickly" was doing too much work there, so the comment now says the unit outright and gives the number: "`fetch-timeout` — in milliseconds — bounds a single stalled request at a minute rather than npm's five-minute default".

reboot/examples/chat-room/frontend/mobile/.tests/maestro_test.sh
rjhuijsman: You read it correctly and the comment was wrong — "then" made it sound like a fallback that kicks in after the retries. It is the opposite: `prefer-offline` applies first. npm normally revalidates a cached package against the registry before reusing it; `prefer-offline` skips that check and takes whatever is already cached, so the pre-cached `.rbt` packages never touch the network at all. Only packages missing from the cache are fetched, and `fetch-timeout`/`fetch-retries` bound just those fetches.

Rewritten to say that, and I reordered the three exports so `prefer-offline` comes first, matching the order the comment now explains them in.

Dockerfile
rjhuijsman: Agreed — done, close to your wording: "When a test runs `rbt dev run` against a fresh state directory it installs these packages, and a warm cache lets it do that without a registry round-trip." No mention of Maestro, so it still reads correctly for whatever else ends up calling `rbt dev run`.

Dockerfile
rjhuijsman: For my block there was no good reason — I copied the condition from the Maestro CLI install just above it, and it was wrong to. These are three pure-JS packages, useful on any architecture that runs `rbt generate`. Removed, so the cache is warmed on arm64 too.

On the Android SDK block above (line 663): that one is deliberate, and for a reason unrelated to the Reboot SDK. It installs `system-images;android-34;google_apis;x86_64` and adds the user to the `kvm` group for the hardware-accelerated emulator; the existing comment at line 653 puts it as "the emulator and its system image are x86_64-only and the test is tagged `requires-linux-x86`". So that restriction is about the Android emulator image, not about anything of ours. Happy to look into an arm64 system image separately if you want the Maestro tests running there, but that is a bigger change than this PR.

.claude/rules/github-prose-no-hard-wrap.md
```diff
@@ -0,0 +1,24 @@
+# Don't hard-wrap prose posted to GitHub
```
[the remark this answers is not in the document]
rjhuijsman: Added a PR description line about why that's here.

Dockerfile
```diff
@@ -710,6 +710,16 @@ RUN if [ "${TARGETARCH}" = "amd64" ]; then \
...
     && rm /tmp/maestro-install.sh; \
     fi
+
+# Warm the npm cache with the packages `rbt generate` installs into a
+# project's `.rbt` directory. These versions must match the ones
+# `_check_or_install_npm_packages` asks for in
+# `reboot/cli/commands/generate.py`. When a test runs `rbt dev run`
+# against a fresh state directory it installs these packages, and a
+# warm cache lets it do that without a registry round-trip.
+RUN npm cache add @bufbuild/protoplugin@1.10.1 \
```
onelxj: Minor but in the file we usually extract the version of some resource to a specific variable and use it from there.

onelxj (approved):

PR #154, opened 2026-09-04
Docs overhaul

rjhuijsman:
Before this PR, the docs presented Reboot as a framework to build _either_ an AI chat app, _or_ a web app.  It also had a "Quick"start that wasn't quick at all, barely mentioned auth, and the whole thing read like a different product than what we sell on the `reboot.dev` home page. It wasn't a good representation of Reboot at all!

This PR reorganizes the docs and (re)writes a lot of content to fix that.

The high-level changes:

- **One product, one argument.** The opening pages carry the `reboot.dev` message and vocabulary, so arriving from the marketing page has a continuous story.
- **One app, several surfaces**: one application can have all of a web frontend, MCP UIs, a React Native app, MCP tools, [...] - it's all just _one_ application.
- **`User`s and auth front and center**: our auth story has matured a lot, and you need to know about it to build a good app.
- **The agent path is the front door.** One guide per agent (Claude/Codex); the hand-write-it Quickstart is gone (but the step-by-step guides remain for readers who want to see every file).
- **Python → Pydantic, TypeScript → Zod.** Protobuf remains only where it is the actual encoding with no alternative API.

**Key decisions** I'd like feedback on:
- React Native is documented, but marked **alpha**. The `reboot.dev` marketing page still says mobile is "Soon".
- Protobuf is mostly removed from the docs, as it is now an implementation detail for ~all users.
- `rbt init` is removed from the docs, because it still scaffolds a protobuf project. We can put it back when we update it to Pydantic.


documentation/docs/surfaces/overview.mdx
```diff
@@ -0,0 +1,107 @@
...
+--react=<dir>` emits one typed React client that the web app, the
+native app, and any [`UI` methods](/surfaces/ui_methods) all import
+from. On the client, sign-in is just `useSignIn()` — on the web and in
+a native app; a chat client signs the person in before any `UI` loads
+— and the generated `useUser()`, which resolves the caller's own
+`User`:
+
+```tsx
+const { user, isLoading } = useUser();
+const { response } = user.useListTodoLists();
```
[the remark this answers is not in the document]
rjhuijsman: Showing `user.useListTodoLists()` is a major part of the goal of this snippet, I don't love pushing that into a sub-component that's not shown in the example (even if there's a comment saying what it does). Is it possible to use the `SignIn` component if still loading, but have `const { response } = user.useListTodoLists()` explicitly?

documentation/docs/surfaces/ai_chat.mdx
```diff
@@ -134,63 +67,105 @@ api = API(
...
+[auto-constructs their `User`](/users/overview) exactly as it would
+for a browser.
+
+This is what makes `User` the natural entry point on this surface in
+particular. The AI's memory of a conversation is its context window,
+which does not survive into the next conversation. State IDs it was
+handed — a `counter_id`, an `order_id` — are gone with it. So:
+
+- Record what the user owns in `User` state.
+- Give `User` a `list_…` method with `mcp=Tool()` so the AI can find
```
onelxj: This `list_... method so AI can find its way back` looks confusing, not sure I got the point.

I guess you mean `list_orders` so later we can find the proper `order_id` to show in another chat, but it sounds like a really specific case rather than generic?

documentation/docs/surfaces/overview.mdx
```diff
@@ -0,0 +1,121 @@
...
+| --- | --- | --- |
+| **Web** | A React app in the browser, talking to your Reboot backend. | [Web apps](/surfaces/web) |
+| **React Native** | A React Native app on iOS and Android, using the same generated hooks as the web app. In alpha. | [React Native apps](/surfaces/react_native) |
+| **AI chat** | Your methods as MCP tools, and React components rendered in the conversation as MCP Apps, inside ChatGPT, Claude, or VS Code. | [AI chat apps](/surfaces/ai_chat) |
+
+## Agents
+
+An AI agent reaches your app over [MCP](/surfaces/ai_chat). Every
+method, `User` methods included, says whether it is a tool:
+`mcp=Tool()` exposes it to the agent, `mcp=None` keeps it out of
```
onelxj: I'd rather hide this technical aspect from that page and leave the explanation to the `ai_chat` page only.

documentation/docs/frontends/overview.mdx
```diff
@@ -0,0 +1,121 @@
...
+**One set of generated hooks, and one way in.** `rbt generate
+--react=<dir>` emits one typed React client that the web app, the
+native app, and any [`UI` methods](/surfaces/ui_methods) all import
+from. On the client, sign-in is just `useSignIn()` — on the web and in
+a native app; a chat client signs the person in before any `UI` loads
+— and the generated `useUser()`, which resolves the caller's own
+`User`:
+
+```tsx
+const Root = () => {
```
onelxj: It is the only code example on that page, should we also extract it from here and put to a specific page to leave overview page be more like mental model of Reboot?

documentation/docs/surfaces/overview.mdx
```diff
@@ -0,0 +1,121 @@
...
+  native app opens a system browser. A chat client runs the flow when
+  the user connects your app.
+- **What the user sees.** A full page, a native screen, or a component
+  embedded in a conversation.
+
+## Serving several at once
+
+Nothing stops one application from doing all of it. The
+[`bank-pydantic`](https://github.com/reboot-dev/reboot-bank-pydantic)
+example (described under [Examples](/get_started/examples#bank-pydantic))
```
onelxj: `described under [Examples](/get_started/examples#bank-pydantic)` looks redundant

documentation/docs/surfaces/overview.mdx
```diff
@@ -0,0 +1,121 @@
...
+does: one backend, one `User` type, and three front ends that all
+import the same generated client.
+
+| Front end | In the example | Docs |
+| --- | --- | --- |
+| Browser app | `frontend/web/` | [Web apps](/surfaces/web) |
+| React Native (Expo) app | `frontend/mobile/` | [React Native apps](/surfaces/react_native) |
+| MCP UIs, one directory per `UI` method | `frontend/mcp/` | [`UI` methods](/surfaces/ui_methods) |
+| Generated client, shared by all three | `frontend/api/` | [Call your API from React](/call/from_react) |
```
onelxj: If a reader will navigate to github and open frontend/api - they'll see the API only, no generated client at all.
Also worth mentioning, shall we just remove that section from there and put to the `bank-pydantic` example readme instead?

documentation/docs/users/providers.mdx
```diff
@@ -0,0 +1,305 @@
...
+All of these live in `reboot.aio.auth.oauth_providers`.
+
+| Provider | Use it when |
+| --- | --- |
+| `Development()` | Local development. Shows a fake account picker; no registration anywhere. |
+| `Anonymous()` | You want no sign-in at all. Every visitor is a new person. |
+| `Google(...)` | Users sign in with Google. |
+| `GitHub(...)` | Users sign in with GitHub. |
+| `Auth0(...)` | You want several sign-in methods, or user management, behind one provider. |
+| `Ory(...)` | Same, on Ory Network or self-hosted Ory. |
```
onelxj: What does `same` relate to? (I guess to `Auth0` but is that explicit enough)?

documentation/docs/concepts.mdx
```diff
@@ -0,0 +1,256 @@
...
+  across clients and restarts.
+
+## What you still write yourself
+
+Reboot is not a no-code tool. You still design your types, write your
+business logic, and build your UI. What it takes off your plate is the
+infrastructure underneath: persistence, transactions, concurrency
+control, reactive fan-out, background execution, sign-in, and per-user
+authorization — the hard problems, solved once at the framework level,
+so neither you nor your coding agent ever has to solve them again.
```
onelxj, line 248: I love that page!

onelxj (approved):

documentation/docs/define/methods.mdx
```diff
@@ -31,15 +31,19 @@ The five kinds of methods in Reboot are:
...
-from the AI). This applies to all state types, including
-`User`. See
-[Creating tools](/learn_more/define/pydantic#creating-tools-for-the-ai)
-for details.
+| `UI` | `UI` methods declare a React component to display in the AI chat interface. The React `App.tsx` is the implementation; there is no backend servicer code. When the AI calls the associated tool, the chat client opens the React app. See [UI methods](/surfaces/ui_methods). |
+
+## Exposing methods to an AI
+
+Every method is callable from your own code, from a
+[web](/surfaces/web) or [React Native](/surfaces/react_native)
```
benh: I don't love the word "surface", I'd rather use frontend or user interface or even just interfaces. It appears as though the word "surface" is only used in the path, which is typical coding agent crap, using two words for one concept. Can we just use `frontends` instead of `surfaces` as the directory name?

documentation/docs/define/methods.mdx
```diff
@@ -31,15 +31,19 @@ The five kinds of methods in Reboot are:
...
+frontend, and from anywhere else you have a client. Whether an **AI**
+may call it is a separate, explicit decision.
+
+When you define your API with [Pydantic](/define/pydantic), every
+method declares its MCP exposure with `mcp=Tool()` (expose it as an
+AI-callable tool) or `mcp=None` (hide it from the AI). There is no
+implicit default, so nothing reaches an AI by accident. This applies
+to all state types, including `User`. See
+[Creating tools](/define/pydantic#creating-tools-for-the-ai) and
+[AI chat apps](/surfaces/ai_chat).
```
benh: I'd love to use this opportunity to just call them MCP UIs instead of AI chat apps and trying to create our own category, even though I thought that might have been the better call in the past.

benh (approved):

documentation/docs/concepts.mdx
rjhuijsman: 🎉

PR #160, opened 2026-09-09
Update Envoy, Istio, and various CLIs

rjhuijsman:
A simple update to many of our dependencies: 
* we move Istio to `1.30.4` (its latest stable)
* we move Envoy to `1.38.4`, which is the closest match to that Istio's Envoy version.
* we update a bunch of CLIs in the `Dockerfile` to their latest versions.

While we're here, de-flake a test that caused a CI failure, and try to harden the `Dockerfile` against networking flakes.

### DIFF
--- a/.github/workflows/reboot_macos_environment.yml
+++ b/.github/workflows/reboot_macos_environment.yml
@@ -143,7 +143,7 @@ jobs:
 
       - name: Install uv
         run: |
-          curl -LsSf https://astral.sh/uv/0.11.13/install.sh | sh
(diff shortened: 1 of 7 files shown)


reboot/plugin/skills/upgrade/migrations/next/envoy-1-38-4.md
rjhuijsman, line 1: Installing Envoy yourself is not a typical deployment path; people use Reboot base images that match the Reboot release. Don't add this migration note.

onelxj (approved):

PR #162, opened 2026-09-10
`state_managers`: fix two read-only transaction participant bugs

rjhuijsman:
This PR fixes two bugs in how read-only transaction participants work in our transaction protocol:
1. A re-sent `Prepare` (by a retrying coordinator) would be told "no such transaction"
2. A read-only participant could be told to commit a transaction it never prepared

Please see the individual commits for details.


rbt/v1alpha1/transactions.proto
```diff
@@ -25,6 +25,17 @@ message PrepareRequest {
...
+  // When true, the coordinator recorded this participant in the
+  // read-only subset of its `Participants` record, which is what
+  // allowed the participant to elide its prepare/commit disk writes,
+  // release its shared lock and forget the transaction on an earlier
+  // `Prepare`. It lets the participant answer a `Prepare` for a
+  // transaction it no longer knows about: having not restarted since
+  // the transaction began, the only way it can have forgotten is by
+  // having already prepared and committed. New coordinators set this
+  // alongside `read_only_aware`; old coordinators leave it false (the
+  // default) and the participant answers `abort = true` as before.
```
rjhuijsman: This is a lot of implementation detail. Let's shorten this dramatically to just the contract: what is the coordinator telling the participant? Something like "The coordinator has recorded this participant as a read-only participant of this transaction."

reboot/aio/state_managers.py
```diff
@@ -5265,6 +5272,15 @@ async def prepare(state_type: StateTypeName, state_ref: StateRef):
...
                             # field and do the disk-writes.
                             read_only_aware=True,
+                            # Tell the participant whether we recorded
+                            # it in that read-only subset, so that if
+                            # this is a retry of a `Prepare` it
+                            # already elided it can answer "prepared"
+                            # rather than "no such transaction". Only
+                            # meaningful alongside `read_only_aware`,
+                            # which is what permits the elision in the
+                            # first place.
```
rjhuijsman: Ditto, too much detail about an implementation that's elsewhere. Only document how we're satisfying the contract, if that needs explaining.

reboot/aio/state_managers.py
```diff
@@ -5681,6 +5704,41 @@ async def Prepare(
...
             if request.abort_via_response:
+                if (
+                    request.read_only and request.read_only_aware and
+                    can_use_restart_detection
+                ):
+                    # This transaction is one a `read_only_aware`
+                    # coordinator recorded as read-only, and the check
+                    # above establishes that this server has not
+                    # restarted since it began. Had we still been
+                    # holding it, the lookup above would have found
```
rjhuijsman: "it" is ambiguous here; I think you mean "this transaction", but please spell it out.

reboot/aio/state_managers.py
```diff
@@ -5681,6 +5704,41 @@ async def Prepare(
...
+                    # memory.
+                    #
+                    # That check is per process: shards are fixed when
+                    # a state manager is constructed, so a shard
+                    # changes owner only via a new process, which the
+                    # recovery timestamp catches. Reassigning a shard
+                    # to an already running server would need a
+                    # per-shard recovery timestamp, or this answer
+                    # would claim a transaction the new owner never
+                    # saw was prepared.
```
rjhuijsman: Make it clearer that "reassigning" is not a thing we currently support, e.g.: "If we implement reassignment of shards into already running servers [...]"

tests/reboot/state_manager_tests.py
```diff
@@ -840,77 +841,347 @@ async def test_transaction_with_idempotency_key_is_exclusive(self):
             assert transaction is not None
             self.assertEqual(transaction.mode, Lock.Mode.EXCLUSIVE)
 
+    def create_grpc_context_mock(
```
benh: These fixes are great, thank you!

But these tests are something I've seen the agent do a lot recently which is instead of testing the implementation they "mimic" the implementation and then test that. This will cause rot very quickly if we change the actual implementation and forget to change these tests. So instead, please ask your robot to use GeneralServicer and mock everything as necessary and write tests that always test the real implementation. Otherwise this looks great!

benh (approved):

tests/reboot/state_manager_tests.py
rjhuijsman: Done; the robot rewrote using `Bank` and `Account` which works too. 👍🏻

PR #168, opened 2026-09-11
Pubsub: let a queue subscribe to the same topic more than once

rjhuijsman:
Before this PR, PubSub's `Topic.Subscribe` appended to `state.queue_ids` without checking whether that queue was already subscribed, so subscribing the same queue to a topic twice left a duplicate entry. `Topic.Broker` fans out over `state.queue_ids`, so the duplicate made it call `Enqueue` on that one queue twice using a single `context`, which Reboot refuses:

```
ValueError: To call 'rbt.std.collections.queue.v1.QueueMethods.Enqueue' of 'receiving-queue' more than once using the same context an idempotency alias or key must be specified
```

The broker then retried with backoff forever, so the topic never drained. Depending on the situation it may be a race — a second `Subscribe` must land before the topic is drained — so this was seen as a flake in `//tests/reboot/std/pubsub/v1:test_pubsub_tests_ts` in [this CI job](https://github.com/reboot-dev/reboot/actions/runs/34603229734/job/103275784303) on an unrelated PR.

This commit makes it so that a second `Subscribe` does nothing.

TESTED: new unit test.


reboot/std/pubsub/v1/pubsub.py
```diff
@@ -111,9 +114,14 @@ async def slice_items(
                 have_items,
             )
 
+            # A topic subscribed to before `Subscribe` deduplicated may
+            # have a repeated queue id persisted. `until` memoizes the
+            # sliced list and `slice_items` has already taken those
+            # items out of state, so the duplicate replays on every
+            # retry; deduplicating here is what lets such a topic drain.
```
benh:
```suggestion
            # In an earlier implementation there was a bug where a topic
            # could be subscribed to by the same queue id more than
            # once. That bug has been fixed, but to handle any `Queue`
            # instances that have a repeated queue id already persisted
            # we also deduplicate here via `dict.fromkeys`.
```

benh (approved):

PR #173, opened 2026-09-14
Coalesce query responses a reactive reader has fallen behind on

rjhuijsman:
Before this PR, if a client was slower to process the results of a reactive read than the server was at producing it (e.g. due to a slow render loop), the client would fall far behind due to large buffers building up in the networking stack. This PR makes it so the speed at which a client _consumes_ results is independent from the speed at which it _processes_ results, coalescing results as needed, so that clients can always consume at full network speed and always show the latest state in their next render loop.

A reactive read has two kinds of "client" that read its responses one at a time: the web client (React, React Native, or the plain-TS `reactively()`), and the server of a reactive reader R1 that transitively calls another reader R2. Before this PR, both could fall behind, and both then had to work through every stale response they had missed, one by one, even though only the latest one mattered. This PR addresses both.

Note that the buffers that would cause a large backlog to build up are in the gRPC library, OS TCP stack, and in Envoy - not in our application code. This PR therefore doesn't remove any buffers (it can't really), it just decreases the odds they fill up. As long as the network can keep up, the clients will keep up.


reboot/templates/reboot_react.ts.j2
```diff
@@ -919,24 +919,37 @@ class {{ client.proto.state_name | to_camel }}Instance {
             // orphans list with a length greater than 0. In this case, we don't
             // want to skip checking the expecteds list just because we have
             // already checked the orphans list.
-            if (
+            //
+            // A single response may carry the idempotency keys of several
+            // expected mutations, both because the server reports every key
+            // since its previous response and because
+            // `reboot_web.reactiveReader()` merges the keys of the responses
+            // that arrived while we were busy above, so we observe each of
```
rjhuijsman: The "because" discussion is implementation details of code that's not here and therefore can change. Comments should discuss only three things:
1. Implementation choices for code that's right adjacent to the comments.
2. Contracts for how other code should expect this code to behave.
3. Contracts for how this code should expect other code to behave.

The "A single response may carry the idempotency keys of several expected mutations" and "Mutations commit in order, so a key later in `expecteds` only shows up once each earlier one has" are good examples of (3). The implementations that make that so are NOT one of these categories.

Add or strengthen a repo-level memory for this.

reboot/templates/reboot_react.ts.j2
```diff
@@ -919,24 +919,37 @@ class {{ client.proto.state_name | to_camel }}Instance {
...
-	          // pass on the response because it might contain new data that should get shown to the
-	          // user (e.g., in a chat room this could be a new message from a different user).
-            else if (response !== undefined && !haveOrphans) {
+            // If we don't have any orphans to observe and we don't have any
+            // expecteds to observe, or at least, the first expecteds _is not
+            // observed_ by this response, then go ahead and pass on the
+            // response because it might contain new data that should get
+            // shown to the user (e.g., in a chat room this could be a new
+            // message from a different user).
+            if (!observedExpected && response !== undefined && !haveOrphans) {
```
rjhuijsman: This implies that if we observed an orphan/expected the response will _not_ contain new data, ever? Is that true? What guarantees that?

reboot/web/index.ts
```diff
@@ -462,6 +418,87 @@ export function reactively<
   return [responses(), setRequest];
 }
 
+// Merges `next`, which arrived after `previous`, into the one response
+// that stands for both. The payload is the latest one either carries,
```
rjhuijsman: "the one response" -> "one response"

PR #174, opened 2026-09-14
`reboot/server`: skip an already-committed participant in `RecoverTransactions`

rjhuijsman:
`RecoverTransactions` iterates a RocksDB snapshot of the `transaction-participant:*` records and, for each, looks the transaction up in memory via `LookupOrBeginTransaction` and then `CHECK`s that it is either `PREPARED` or `STARTED`. The snapshot and the in-memory lookup are not taken atomically, and nothing stops a `TransactionParticipantCommit` (or `Abort`) from landing in between: several servers' `Recover` RPCs run concurrently at start-up. Committing deletes the participant record and then, via `DeleteTransaction`, the in-memory transaction, so a stale snapshot during recovery might find a transaction that is already `COMMITTED` (or `ROLLEDBACK`), which tripped the `CHECK` and took the whole process down:

```
F… database.cc:3744] Check failed: (**txn)->GetState() == rocksdb::Transaction::STARTED (4 vs. 0)
```

This was spotted on macOS arm64 CI twice, both in tests that do `down()` then `up()` right after a transactional call:
- https://github.com/reboot-dev/reboot/actions/runs/34852764130/job/104038177834 (`//tests/reboot:preload_tests_py`, two database threads hit the `CHECK` 100 µs apart)
- https://github.com/reboot-dev/reboot/actions/runs/34834498052/job/103946421722 (`//tests/reboot/std/ciphertext/v1:ciphertext_tests_py`)

The same race can also produce a second, quieter failure mode: if `DeleteTransaction` has already run by the time recovery reaches the stale record, `LookupOrBeginTransaction` finds nothing in memory and begins a fresh, empty transaction for the already-deleted record, reporting it as an unprepared participant. Its coordinator can never commit that (`Commit()` on an unprepared transaction with `skip_prepare = false` is `TxnNotPrepared`), so the state stays stuck behind the participant's exclusive lock forever.

This commit fixes the race. `RecoverTransactions` now looks the transaction up in memory itself (`LookupTransaction`) and, when it is absent, begins one only if the record is still in the database rather than merely in the snapshot (a point read, done after the in-memory lookup because the record is always deleted before the transaction is, so a transaction that is gone from memory has taken its record with it). A transaction found `COMMITTED` or `ROLLEDBACK` is skipped: our borrow of it is all that is holding off its `DeleteTransaction`, and there is nothing left to recover. The `CHECK` remains for every other unexpected state (`AWAITING_*` after a failed prepare/commit), and the transaction's name is still checked against the record, as `LookupOrBeginTransaction` did before.

TESTED: new unit tests reproduce the interleaving that used to result in these races.


benh (approved):
🙏

PR #185, opened 2026-09-18
React+Web: reactive readers shouldn't retry definitive errors

rjhuijsman:
Reboot's documented retry policy ([Automatic retries](https://github.com/reboot-dev/reboot/blob/main/documentation/docs/call/from_react.mdx#automatic-retries)) is that "an error is considered unretryable if it originates from the application", and a unary call from the generated React client follows it: it retries HTTP 502, 503 and 499 and failed fetches indefinitely, and returns every other error as `aborted`.

Before this PR, a reactive reader hook made no such distinction. Any error raised by the reader, a denied authorizer, or `StateNotConstructed` was retried every 1-3 seconds ~forever.

This commit brings the reactive reads in line with unary reads: a failed connection and a `Status` with a retryable code (`UNAVAILABLE` or `CANCELLED`, the gRPC codes behind HTTP 503 and 499) are retried. Any other `Status` is final: the reader surfaces it as `aborted` with `isLoading` `false`. It will read again only once a mutation on the same state is made through this client, or when the hook is remounted with different inputs.

There are two commits: the first makes the fix for React, the second for plain-web.

The "reads again after a mutation" part goes one step beyond a literal port of the unary policy, and is there because e.g. the `chat-room` example (and the docs' "Reading state" section) mount a reader before the state exists: "the state will be constructed when we make the first `.writer` call". Under a literal port, that reader would show `aborted: StateNotConstructed` forever after the first `send`. With the wake-on-mutation, the reader reconnects right after the constructing mutation flushes, so the pattern keeps working, without polling.

Review request: please consider extra carefully whether we're happy with...
1. **A reader whose declared error is expected to clear because of _other_ clients' writes no longer picks that up.** E.g. a reader raising `NoWinnerYet` until another player ends the game, or a component that mounts a reader on a state the backend constructs later. Before these commit, the infinite retries meant that they'd eventually discover their error had cleared; now they stay `aborted` until this client mutates the state or the hook remounts. The migration fragment tells upgrading apps to return the condition in the response instead of raising it, which keeps the stream open and reactive, but that's not an option for built-in definitive errors like `StateNotConstructed`.
2. **`Unknown`/`Internal` (an unexpected exception in a reader) are final too**, as they are for unary calls. This wasn't our intuition for either unary or reactive when we talked about it, but it seems better to be in line with unary calls than to invent a different semantic.

TESTED: new unit test.


reboot/react/index.tsx
```diff
@@ -888,6 +888,10 @@ export interface Mutate {
 export interface Reader<ResponseType> {
   abortController: AbortController;
   event: Event;
+  // Whether the reader is (re)connecting to the server. It is `false`
+  // while a stream of responses is open and while the reader has
+  // settled on a `status` that a retry would only reproduce.
```
rjhuijsman: "would only reproduce"? I don't understand this English, "reproduce" is something I associate with bugs?

reboot/templates/reboot_react.ts.j2
```diff
@@ -811,6 +811,15 @@ class {{ client.proto.state_name | to_camel }}Instance {
...
     let orphans: Expected[] = [];
 
+    // Set once a mutation has been made on this state since the
+    // current attempt to read started, and the event a reader that
+    // settled on an error waits on for exactly that. A mutation may
+    // change the state such that the reader now succeeds, e.g., a
+    // constructor after `StateNotConstructed`, so it is what makes
+    // reading again worthwhile.
+    let mutated = false;
+    let awaitingMutation: reboot_api.Event | undefined = undefined;
```
rjhuijsman: Is this the industry standard on how to name events? Doesn't that lead to me having to say "I'm awaiting awaitingMutation"?

reboot/web/index.ts
```diff
@@ -450,11 +469,45 @@ export function reactively<
...
+          // The server answered, and its answer is an error, so
+          // reading again would get the same answer. A stale session
+          // is the one such answer a renewal may change, so ask for
+          // one before surfacing it, as a unary call does.
+          if (
+            e.code === StatusCode.UNAUTHENTICATED &&
+            onUnauthenticated !== undefined &&
+            !didRefresh
+          ) {
+            didRefresh = true;
```
rjhuijsman, line 511: Should a successful request after this set `didRefresh` back to `false`, so another stale session later is refreshed again?

PR #186, opened 2026-09-18
Skills and dashboard fixes and improvements

onelxj:
See commits.


reboot/plugin/skills/python/references/state-collections.md
```diff
@@ -270,11 +272,10 @@ class UserState(Model):
 
 class UserServicer(User.Servicer):
 
-    async def create(self, context: WriterContext) -> None:
-        if context.constructor:
-            self.state.profile_id = str(uuid4())
-            self.state.drafts_index_id = str(uuid4())
-            self.state.inbox_queue_id = str(uuid4())
+    async def create(self, context: TransactionContext) -> None:
+        self.state.profile_id = str(uuid4())
```
onelxj, line 276: why did you drop ` if context.constructor:`

reboot/plugin/lib/install_node.sh
```diff
@@ -1,15 +1,21 @@
...
 
-# Pinned Node.js version. To bump, change here and update the
+# Pinned Node.js major version; the newest release of it is
+# installed, so the tools a project pulls in (e.g. Vite) find the
+# minor versions they require. To bump, change here and update the
 # README.
-NODE_VERSION="22.11.0"
+NODE_MAJOR="22"
+
+# How long a resolved release is used before looking for a newer
```
onelxj: Let's pick the latest every time instead.

reboot/plugin/skills/web-app/references/react-client.md
```diff
@@ -15,28 +15,32 @@ MCP-host-specific and do not apply to a web app.
...
+`react`, `react-dom`, and `devDependencies` (Vite, its React plugin,
+TypeScript, the `@types/*`) at the versions it writes; they are made
+to work together. Add the two Reboot packages, pinned to the same
+version as the backend's `reboot` dependency, `@bufbuild/protobuf`
+at the exact version `@reboot-dev/reboot-react` names as its peer
+dependency, and `zod`:
 
 ```json
 {
-  "name": "<app>-web",
```
onelxj, line 23: why did we remove all of that info rfrom the `package.json`?

reboot/plugin/skills/python/references/state-collections.md
```diff
@@ -196,13 +196,15 @@ class UserState(Model):
...
-            self.state.people_index_id = str(uuid4())
+    # The auto-constructed `User`'s constructor, a `Transaction` the
+    # framework calls once, when the user first signs in. It is not
+    # declared in the API; override it to initialize the state.
+    async def create(self, context: TransactionContext) -> None:
+        # Just allocate the ID — the OrderedMap is constructed
+        # implicitly on the first `insert`. Until then, reading it
+        # aborts with `StateNotConstructed`; see
+        # `stdlib-ordered-map.md`.
+        self.state.people_index_id = str(uuid4())
```
benh, line 207: Why don't we have the `context.constructor` anymore? And what do you mean it isn't declared on the API? We definitely show it in the dashboard. Are you 100% sure that nobody outside can call it even via an HTTP call even if we don't generate code for calling it? I would think we'd still want `context.constructor` unless we are somewhere keeping it from being called.

reboot/plugin/skills/python/references/state-collections.md
```diff
@@ -270,11 +272,10 @@ class UserState(Model):
 
 class UserServicer(User.Servicer):
 
-    async def create(self, context: WriterContext) -> None:
-        if context.constructor:
```
benh, line 274: Can we really drop this?

reboot/plugin/skills/python/references/patterns-common-gotchas.md
```diff
@@ -287,5 +287,19 @@ with `await rbt.create_external_context_as(name, user_id)` is enough
...
+
+To give a new `User` initial state (e.g. allocate the id of an
+`OrderedMap` index, `state-collections.md`), override its
+constructor on the servicer; it is not declared in the API:
+
+```python
+class UserServicer(User.Servicer):
+
+    async def create(self, context: TransactionContext) -> None:
+        self.state.todos_index_id = str(uuid4())
```
benh, line 301: Can this really not be called more than once!?

reboot/plugin/skills/python/references/lifecycle-project-setup.md
```diff
@@ -173,6 +173,12 @@ ignore_missing_imports = True
 [mypy-grpc_status.*]
 ignore_missing_imports = True
 
+# `from uuid7 import create as uuid7` (time-ordered keys) comes from
+# the `uuid7-standard` package, which `reboot` depends on; it ships
+# no type information.
+[mypy-uuid7]
+ignore_missing_imports = True
```
benh, line 180: Isn't there a reason that we have our own uuid7 implementation internally? Should we actually be dropping all external uuid7 calls and using that exclusively? Then update the skills to say use our uuid7 because <fill in the reason>?

reboot/plugin/skills/python/references/state-collections.md
```diff
@@ -324,6 +324,17 @@ form runs:
    convention has no such moment; the related actor springs into
    existence on whichever method happens to touch it first.
 
+### Deleting an Entity
```
benh, line 327: What is an entity?

reboot/plugin/skills/python/references/lifecycle-project-setup.md
```diff
@@ -108,9 +110,11 @@ tests records every browser scenario (see `testing-web-app.md`).
 # Reboot dev-server state.
 .rbt/
 
-# Generated code; recreated by `rbt generate`.
+# Generated code; recreated by `rbt generate`. `frontend/api/` is
+# an MCP UI's React client, `web/src/api/` a web app's.
 backend/api/
 frontend/api/
+web/src/api/
```
benh, line 117: Wait what? I thought that you had changed it so that everything "frontend" is in `frontend`, including `web` and `mobile`? This seems like maybe we need to do a whole pass in the skills, docs, and examples and make sure it is all consistent.

reboot/plugin/skills/python/references/testing-web-app.md
```diff
@@ -261,6 +261,18 @@ tests/opening_accounts.recordings/
...
+  told otherwise, so an app that follows the OS theme is recorded
+  (and asserted on) in its light theme only. To run a test module's
+  scenarios in dark mode, override pytest-playwright's
+  `browser_context_args` fixture in it:
+
+  ```python
+  @pytest.fixture
+  def browser_context_args(browser_context_args: dict) -> dict:
+      return {**browser_context_args, 'color_scheme': 'dark'}
+  ```
```
benh, line 275: Instead lets force a suffix on the "open the web app" sentence to be "opens the web app in light mode" or "opens the web app in dark mode". That way a human reading the scenario never needs to know this rule, they always read exactly what it says!

reboot/aio/auth/consent_page.html.j2
```diff
@@ -4,6 +4,7 @@
 <meta charset="utf-8">
 <meta name="viewport" content="width=device-width, initial-scale=1">
 <title>Authorize application</title>
+<link rel="icon" type="image/svg+xml" href="/__/rootpage/favicon.svg">
```
benh, line 7: Why is the favicon.svg always there?

reboot/aio/auth/development_login_page.html.j2
```diff
@@ -4,6 +4,7 @@
 <meta charset="utf-8">
 <meta name="viewport" content="width=device-width, initial-scale=1">
 <title>Who do you want to be?</title>
+<link rel="icon" type="image/svg+xml" href="/__/rootpage/favicon.svg">
```
benh, line 7: So this one is definitely for the Reboot development picker so the Reboot favicon makes sense there ... but I wonder for the other one if it is for their app should we really use the Reboot favicon.svg? Maybe that should be something that needs to be injected some how?

reboot/plugin/skills/python/references/testing-features.md
```diff
@@ -216,7 +216,10 @@ And the resulting `account_id` is saved as "alice account id"
 - **A saved name is a quoted string** and may have spaces, `"first account id"`, and is said back as `<first account id>`: in a
   state id (`of "<account id>"`), a user id, a bearer token, or a
   property value (`amount=<balance>`). A `<name>` inside a JSON
-  string stays literal. A `Scenario Outline`'s columns are said the
+  string stays literal, so in a value write it bare:
+  `todo_id=<errand>` passes the saved id, while `todo_id="<errand>"`
```
benh, line 220: Please look into this as discussed synchronously.

reboot/dashboard/web/src/callgraph.ts
```diff
@@ -49,8 +49,16 @@ export interface GraphMethod {
   factory: boolean;
   calls: GraphCall[];
   runs: GraphRun[];
+  // Whether Reboot adds the method to an auto-constructed state type
+  // and calls it itself as a user signs in, rather than the API file
+  // declaring it; see `INJECTED_METHODS`.
+  injected?: boolean;
```
benh, line 55: Wait what? `injected` sounds very generic but the comment is SUPER specific to "user sign in". And what is the semantics we want here? I definitely want to see what `create` calls since it can be implemented by a servicer. Maybe the right solution here is that we should require `create` to be on the `User` API? Think about it and then lets talk.

reboot/plugin/skills/web-app/SKILL.md
```diff
@@ -575,6 +575,20 @@ Key differences from a `mcp-ui` layout:
...
     `web/src/api/**/*_rbt_react.ts` to check them — it is tens of
     thousands of lines that then ride along on every later turn.
+
+    How the app **looks** is not Reboot's concern and this skill
+    says nothing about it. Before writing any page, load Anthropic's
+    `frontend-design` skill and follow it: it picks a visual
+    direction for this app and holds the page to a modern quality
+    bar. If the skill is not available, say so once and suggest the
+    user install it with
+    `/plugin install frontend-design@claude-plugins-official`, then
```
benh, line 585: Does Codex also load this one? Is it considered the best?

benh (changes requested):

